Issue #38, opened 2016-04-29
GUI tournament options incomplete


rwst:
At the moment there seems no way to specify arbitrary options for a new tournament. The dialog for example does not allow specification of a tournament book. This could be resolved by a field that enables textual input of any supported option. An alternative would be to have a text file with default UCI config options (for both gui/cli) or global GUI data field with such content.

ilaripih:
This is a known issue - the whole GUI side is still very much incomplete.

ilaripih:
Opening books can be used now and otherwise this is a duplicate of #83.

Issue #30, opened 2016-03-26
Usage of obsolete Qt functions


ilaripih:
We're using these obsolete Qt functions in quite a few places: http://doc.qt.io/qt-5/qtalgorithms-obsolete.html
They can and should be easily replaced with std counterparts.

ilaripih:
Also, Qt now has a Q_UNREACHABLE macro which is a better way to flag unreachable branches than qFatal("Unreachable").

artoj:
I think we should enable c++11 while we transition to those std counterparts and document c++11 as requirement.

ilaripih:
Yes, I think C++11 is the way to go. Are there any mainstream Linux distributions that don't ship a C++11 enabled GCC compiler in their standard repositories? And Clang probably handles C++11 just fine?

ilaripih:
I just added C++11 as a requirement, was easy enough.

artoj:
GCC / LLVM seems pretty good. Debian stable has GCC 4.9 by default which has full support for C++11: [GCC C++11 support table](https://gcc.gnu.org/projects/cxx-status.html#cxx11)

For MS VC++ you need 2015 to use all the features of C++11: [MSVC C++11 support table](https://msdn.microsoft.com/en-us/library/hh567368.aspx)

We can make the jump now and start using C++11.

ilaripih:
At least all Travis CI builds (both GCC and Clang) are fine, and my Windows/MinGW tests also worked fine. I'm interested in using at least the following C++11 features:

* New signals and slots syntax where it makes sense: http://doc.qt.io/qt-5/signalsandslots.html
* Lambda expressions for short slot functions, for cases where we would've previously created a private slot that is only used in a single connection.
* The "auto" type, which will save us a lot of typing

ilaripih:
Obsolete <QtAlgorithms> functions are now gone.

artoj:
Visual C++ 2013 seems to support auto and lambda expressions.

Given that we don't have so many unit tests I think C++11 should be primarily used in new code and older code updated for the parts where it makes sense such as one-liner slots and such.

ilaripih:
I've been thinking of putting more effort into having decent unit test coverage at least in the library code. If I do non-trivial changes to old code, I'll make sure that change is covered by a unit test, or at least test it thoroughly manually.

artoj:
[...]

PR #39, opened 2016-05-07
new tournament dialog: add book options; part of #38

rwst:


projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,39 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
 	t->setGamesPerEncounter(ui->m_gamesPerEncounterSpin->value());
 	t->setRoundMultiplier(ui->m_roundsSpin->value());
 
+	QString bookFileName = ui->m_bookFileEdit->text();
+	if (not bookFileName.isEmpty())
```
ilaripih: We're using `!` instead of `not`

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,39 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
...
+	{
+		OpeningSuite::Format format = OpeningSuite::PgnFormat;
```
ilaripih: This section could be shorter by 4 lines:

```
OpeningSuite::Format format = OpeningSuite::PgnFormat;
if (bookFileName.endsWith(".epd"))
	format = OpeningSuite::EpdFormat;
```

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,39 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
...
+		else
+			{} // assume the user knows what he is doing
+
+		OpeningSuite::Order order = OpeningSuite::SequentialOrder;
+		if (ui->m_seqOrderRadio->isChecked())
+			order = OpeningSuite::SequentialOrder;
+		else if (ui->m_randomOrderRadio->isChecked())
+			order = OpeningSuite::RandomOrder;
+		else
+			return 0;
```
ilaripih: Can the code even reach this line? If it can, the UI logic should probably be reworked so that one of the radio buttons must always be checked.

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,39 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
...
+
+		OpeningSuite* suite = new OpeningSuite(ui->m_bookFileEdit->text(),
+		                         format, order, 0);
+		bool ok = suite->initialize();
+		if (ok)
+			t->setOpeningSuite(suite);
+		else
+			return 0;
```
ilaripih: On successful suite initialization the Tournament object takes ownership of the suite. But if initialization fails, you have to delete the suite manually to prevent a memory leak.

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,39 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
...
+		if (ui->m_seqOrderRadio->isChecked())
+			order = OpeningSuite::SequentialOrder;
+		else if (ui->m_randomOrderRadio->isChecked())
+			order = OpeningSuite::RandomOrder;
+		else
+			return 0;
+
+		OpeningSuite* suite = new OpeningSuite(ui->m_bookFileEdit->text(),
+		                         format, order, 0);
+		bool ok = suite->initialize();
```
ilaripih: No need to create a new `ok` variable for this.

ilaripih:
Thanks for the pull request! The changes are definitely useful and needed, but I suggested a few improvements and fixes.

Also, please consider the big picture where we have opening suites (PGN and EPD files that set a starting position for the game) and opening books (Polyglot books that can be set individually for each engine). It's even possible to set an opening suite for the games and give players opening books at the same time.

So could you please change any references to "books" or "opening books" to "opening suites" in both the code and the UI?

rwst:
Thanks for the review. It was fun to do app devel after some years of computer algebra. Qt5 is a marvel.

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -83,6 +84,9 @@ NewTournamentDialog::NewTournamentDialog(EngineManager* engineManager,
 	connect(ui->m_browsePgnoutBtn, SIGNAL(clicked()),
 		this, SLOT(browsePgnout()));
 
+	connect(ui->m_browseFilesBtn, SIGNAL(clicked()),
```
ilaripih: I think "m_browseFilesBtn" and "browseFiles()" are a bit too generic as names when they are only meant to browse opening suites. Maybe "m_browseOpeningSuiteBtn" and "browseOpeningSuite()"?

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,35 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
 	t->setGamesPerEncounter(ui->m_gamesPerEncounterSpin->value());
 	t->setRoundMultiplier(ui->m_roundsSpin->value());
 
+	const QString& fileName = ui->m_suiteFileEdit->text();
```
ilaripih: The const reference doesn't really make sense here. `ui->m_suiteFileEdit->text()` returns a `QString` object so this line should be:

`QString fileName = ui->m_suiteFileEdit->text();`

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,35 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
...
+	if (!fileName.isEmpty())
+	{
+		OpeningSuite::Format format = OpeningSuite::PgnFormat;
```
ilaripih: Now the opening suite format is always PGN

projects/gui/src/newtournamentdialog.cpp
```diff
@@ -240,6 +254,35 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
...
+		if (suite->initialize())
+			t->setOpeningSuite(suite);
+		else
+		{
+			delete suite;
+			return 0;
+		}
+	}
+
+	if (ui->m_repeatCheckBox->isChecked())
```
ilaripih: This whole thing can be squeezed to a single line:

`t->setOpeningRepetition(ui->m_repeatCheckBox->isChecked());`

ilaripih:
Qt5 is indeed very nice to develop with, especially with C++11 which we'll be using more in the future: https://github.com/cutechess/cutechess/issues/30#issuecomment-202872657

I added a couple of comments about your latest commit.

ilaripih:
The "New tournament" dialog's 966 pixel height is too much many screens so the layout needs to be different. I can tweak it tomorrow. Otherwise, things look good.

PR #40, opened 2016-05-08
apply clang-tidy modernize-use-nullptr on some files

rwst:
This is for demonstration purposes, the result of:
```
ralf@ark:~/cutechess> clang-tidy -p compile_commands.json projects/gui/src/engine* -checks=modernize-use-nullptr -fix
```
I used newest svn of llvm/clang/clang-extras and the Bear utility to generate the json.

`nullptr` is one of the useful C++11 modernizations because of clarity and type-safety.

### DIFF
--- a/projects/gui/src/engineconfigurationdlg.cpp
+++ b/projects/gui/src/engineconfigurationdlg.cpp
@@ -38,7 +38,7 @@ EngineConfigurationDialog::EngineConfigurationDialog(
 	EngineConfigurationDialog::DialogMode mode, QWidget* parent)
 	: QDialog(parent),
 	  m_engineOptionModel(new EngineOptionModel(this)),
-	  m_engine(0),
+	  m_engine(nullptr),
 	  ui(new Ui::EngineConfigurationDialog)
 {
 	ui->setupUi(this);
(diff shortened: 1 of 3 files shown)


rwst:
Please see also other modernize* branches at my repo and tell me how a full modernization should be submitted.

ilaripih:
`nullptr` definitely adds clarity in some places. But Qt also has the Q_NULLPTR macro which expands to `nullptr` if C++11 is enabled, even on the upcoming Qt 5.7 which requires C++11.

We'll have to discuss this with Arto. We don't really any clear guidelines for contributing code to Cute Chess, but this is a really nice problem to have!

rwst:
`nullptr` could be easily converted to `Q_NULLPTR`.

artoj:
Since we require C++11 I think **nullptr** is the one we want to use for new code.

rwst:
> Since we require C++11 I think nullptr is the one we want to use for new code.

Note also gcc-6.1 defaults to C++14: https://gcc.gnu.org/gcc-6/changes.html
I'll close this demo PR and submit one or several full ones.

PR #41, opened 2016-05-08
range-based for loop suggest by clang-tidy

rwst:
This is the only change suggested regarding range-based for loops in `projects/gui/src/*` so I think this is fine as PR.

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -341,9 +341,9 @@ void MainWindow::setCurrentGame(const TabData& gameData)
 	if (gameData.game == m_game && m_game != 0)
 		return;
 
-	for (int i = 0; i < 2; i++)
+	for (auto & m_player : m_players)
 	{
-		ChessPlayer* player(m_players[i]);
+		ChessPlayer* player(m_player);
 		if (player != 0)
 		{
 			disconnect(player, 0, m_engineDebugLog, 0);
--- a/projects/lib/src/chessgame.cpp
+++ b/projects/lib/src/chessgame.cpp
@@ -142,10 +142,10 @@ void ChessGame::stop(bool emitMoveChanged)
 void ChessGame::finish()
 {
 	disconnect(this, SIGNAL(playersReady()), this, SLOT(finish()));
-	for (int i = 0; i < 2; i++)
+	for (auto & i : m_player)
 	{
-		if (m_player[i] != 0)
-			m_player[i]->disconnect(this);
+		if (i != 0)
+			i->disconnect(this);
 	}
 
 	emit finished(this);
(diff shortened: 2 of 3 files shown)


projects/gui/src/mainwindow.cpp
```diff
@@ -341,9 +341,9 @@ void MainWindow::setCurrentGame(const TabData& gameData)
 	if (gameData.game == m_game && m_game != 0)
 		return;
 
-	for (int i = 0; i < 2; i++)
+	for (auto & m_player : m_players)
```
artoj: The **m_** prefix is used for member variables in classes. Since we're iterating over **m_players** the variable name should be **player**.

projects/lib/src/chessgame.cpp
```diff
@@ -142,21 +142,21 @@ void ChessGame::stop(bool emitMoveChanged)
 void ChessGame::finish()
 {
 	disconnect(this, SIGNAL(playersReady()), this, SLOT(finish()));
-	for (int i = 0; i < 2; i++)
+	for (auto & i : m_player)
 	{
```
artoj: Since we're not iterating over indexes anymore I think it would make more sense to have **player** as the auto-variable and maybe change **m_player** -> **m_players** so it reads better. @ilaripih ?

projects/lib/src/chessgame.cpp
```diff
@@ -142,21 +142,21 @@ void ChessGame::stop(bool emitMoveChanged)
...
-		if (m_player[i] != 0)
-			m_player[i]->disconnect(this);
+		if (i != 0)
```
artoj: I think new code should use **nullptr** from now on instead of zero.

rwst:
Thanks. I agree with the comments but should it be nullptr or Q_NULLPTR (see the other demo PR).

projects/lib/src/chessgame.cpp
ilaripih: @artoj I agree completely.

projects/gui/src/mainwindow.cpp
```diff
@@ -341,9 +341,9 @@ void MainWindow::setCurrentGame(const TabData& gameData)
...
 	{
-		ChessPlayer* player(m_players[i]);
+		ChessPlayer* player(m_player);
```
ilaripih: This line could be removed completely now, and the "for" statement could just be `for (auto & player : m_players)`

projects/lib/src/chessgame.cpp
```diff
@@ -528,10 +528,9 @@ void ChessGame::syncPlayers()
 {
 	bool ready = true;
 
-	for (int i = 0; i < 2; i++)
+	for (auto player : m_player)
 	{
-		ChessPlayer* player = m_player[i];
-		Q_ASSERT(player != 0);
+			Q_ASSERT(player != 0);
```
ilaripih: Where did the extra indentation come from?

rwst:
It looks like some of the tools are buggy with whitespace (see also the auto branch of my repo). I'll finish the nullptr branches which look okay, and then checkout and try the last stable version of clang for the loop enhancement. Or file a bug report and repair the output manually.

PR #45, opened 2016-05-14
nullptr in gui/src/boardview/

rwst:


ilaripih:
Looks good, I'll merge.

PR #47, opened 2016-05-14
nullptr in gui/src/mainwindow.*

rwst:


projects/gui/src/mainwindow.cpp
```diff
@@ -338,28 +338,28 @@ void MainWindow::destroyGame(ChessGame* game)
...
 	for (int i = 0; i < 2; i++)
 	{
 		ChessPlayer* player(m_players[i]);
-		if (player != 0)
+		if (player != nullptr)
 		{
-			disconnect(player, 0, m_engineDebugLog, 0);
-			disconnect(player, 0, m_chessClock[0], 0);
-			disconnect(player, 0, m_chessClock[1], 0);
+			disconnect(player, nullptr, m_engineDebugLog, nullptr);
```
ilaripih: I'm not so sure that we should use **nullptr** with **disconnect**: http://doc.qt.io/qt-5/qobject.html#disconnect

Even though everything else in the Qt API documentation has been converted to use **Q_NULLPTR** instead of 0, the documentation for **disconnect** still says that one should use the value 0 may be used as a wildcard.

PR #50, opened 2016-05-14
nullptr in lib/src/board/

rwst:


projects/lib/src/board/standardboard.cpp
```diff
@@ -1084,7 +1084,7 @@ Result StandardBoard::tablebaseResult(unsigned int* dtm) const
 		}
 	}
 
-	GaviotaTablebase::Castling castling = 0;
```
ilaripih: Castling is a bitflag type (int), not a pointer. So we should keep using 0.

artoj:
Can you squash the revert commit to the 2afd54d commit and re-push it?

PR #151, opened 2016-09-21
Variants1

alwey:
For UCI engines variants often are selected by setting check options _UCI_VariantName_ like UCI_Chess960, UCI_Crazyhouse, UCI_Horde etc. (Using prefix UCI_ for all of this is a bit unlucky).

Another way is to use combo option "UCI_Variant" with a list of supported options.   
Some engines support dozens of variants, e.g. Sjaak II. This engine also uses the UCI_Variant combo box option.

This patch adds support of combo option UCI_Variant.

### DIFF
--- a/cutechess.pro
+++ b/cutechess.pro
@@ -1,6 +1,6 @@
-contains(QT_VERSION, ^4\\..*|^5\\.[0-1]\\..*) {
+contains(QT_VERSION, ^4\\..*|^5\\.[0-2]\\..*) {
     message("Cannot build Cute Chess with Qt version $${QT_VERSION}.")
-    error("Qt version 5.2 or later is required.")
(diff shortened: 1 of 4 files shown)


ilaripih:
Thanks a lot for this pull request! I actually just rewrote and committed the whole thing because it was a bit more complicated than it first seemed:

* An engine may mix `UCI_VariantName` boolean options with `UCI_Variant` combo option, and the boolean option of the previous game should be reset to false when starting a new game with a different variant that may use `UCI_Variant`.
* The `variantFromUci` and `variantToUci` functions had to be adapted to the `UCI_Variant` option.
* `UCI_Variant` should not be configurable or visible to the user in the engine configuration dialog

PR #166, opened 2016-10-03
Berolina 1

alwey:
This is a proposal to support variants with different pawn movements, like the Berolina chess variant.

Initially I intended only to override the pawn move generation of standard board. However, the legality and in-check handling and the support for reading and writing moves and positions from and to files had to be done in base classes.

This version was tested with Sjaak II 1.3.1a and with Fairy-Max 4.8Q. Files were also transferred from and to xboard versions 4.7.3 and 4.9.1. All perft tests of the cutechess berolina variant were identical with the corresponding Sjaak results.

I hope this is useful.

### DIFF
--- a/cutechess.pro
+++ b/cutechess.pro
@@ -1,6 +1,6 @@
-contains(QT_VERSION, ^4\\..*|^5\\.[0-1]\\..*) {
+contains(QT_VERSION, ^4\\..*|^5\\.[0-2]\\..*) {
     message("Cannot build Cute Chess with Qt version $${QT_VERSION}.")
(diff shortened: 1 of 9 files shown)


ilaripih:
Thanks! I'll do a more thorough code review tomorrow. But the first thing I noticed was the Qt version requirement and then restoring it. I'd ask you to squash both commits into a single commit so that the Qt version changes disappear from the pull request.

alwey:
You are welcome. Now I removed a duplicate test case and squashed the commits

alwey:
#167

PR #167, opened 2016-10-03
begin support for Berolina

alwey:
removed a duplicate test case and squashed the commits

### DIFF
--- a/projects/lib/src/board/berolinaboard.cpp
+++ b/projects/lib/src/board/berolinaboard.cpp
@@ -1,0 +1,45 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "berolinaboard.h"
+
+namespace Chess {
+
+BerolinaBoard::BerolinaBoard()
+	: StandardBoard()
+{
+ /*
+  * Berolina pawns move one square diagonally forward and they capture
+  * one square straight ahead.
+  */
+	m_pawnSteps.clear();
+	m_pawnSteps += {FreeStep, -1};
+	m_pawnSteps += {CaptureStep, 0};
+	m_pawnSteps += {FreeStep, 1};
+}
+
+Board* BerolinaBoard::copy() const
+{
+	return new BerolinaBoard(*this);
+}
+
+QString BerolinaBoard::variant() const
+{
+	return "berolina";
+}
+
+} // namespace Chess
--- a/projects/lib/src/board/board.cpp
+++ b/projects/lib/src/board/board.cpp
@@ -319,14 +319,15 @@ QString Board::moveString(const Move& move, MoveNotation notation)
 	return lanMoveString(move);
 }
 
-Move Board::moveFromLanString(const QString& str)
+Move Board::moveFromLanString(const QString& istr)
 {
+	QString str(istr);
+	str.remove(QRegExp("[x+#!?]"));
 	int len = str.length();
 	if (len < 4)
 		return Move();
 
 	Piece promotion;
-
 	int drop = str.indexOf('@');
 	if (drop > 0)
 	{
--- a/projects/lib/src/board/westernboard.cpp
+++ b/projects/lib/src/board/westernboard.cpp
@@ -28,6 +28,7 @@ WesternBoard::WesternBoard(WesternZobrist* zobrist)
 	  m_arwidth(0),
 	  m_sign(1),
 	  m_enpassantSquare(0),
+	  m_enpassantTarget(0),
 	  m_reversibleMoveCount(0),
 	  m_kingCanCapture(true),
 	  m_zobrist(zobrist)
@@ -38,6 +39,10 @@ WesternBoard::WesternBoard(WesternZobrist* zobrist)
 	setPieceType(Rook, tr("rook"), "R", RookMovement);
 	setPieceType(Queen, tr("queen"), "Q", BishopMovement | RookMovement);
 	setPieceType(King, tr("king"), "K");
+
+	m_pawnSteps += {CaptureStep, -1};
+	m_pawnSteps += {FreeStep, 0};
+	m_pawnSteps += {CaptureStep, 1};
 }
 
 int WesternBoard::width() const
@@ -94,6 +99,13 @@ void WesternBoard::vInitialize()
 	m_rookOffsets[1] = -1;
 	m_rookOffsets[2] = 1;
 	m_rookOffsets[3] = m_arwidth;
+
+	m_pawnAmbiguous = (pawnAmbiguity(FreeStep) > 1);
+}
+
+inline int WesternBoard::pawnPushOffset(const PawnStep& ps, int sign) const
+{
+	return sign * ps.file - sign * m_arwidth * 1;
 }
 
 int WesternBoard::captureType(const Move& move) const
@@ -165,6 +177,11 @@ QString WesternBoard::sanMoveString(const Move& move)
 
 	if (piece.type() == Pawn)
 	{
+		if (m_pawnAmbiguous)
+		{
+			needFile = true;
+			needRank = true; // for Xboard-compatibility
+		}
 		if (target == m_enpassantSquare)
 			capture = Piece(side.opposite(), Pawn);
 		if (capture.isValid())
@@ -297,10 +314,11 @@ Move WesternBoard::moveFromSanString(const QString& str)
 	QString::const_iterator it = mstr.begin();
 
 	// A SAN move can't start with the capture mark, and
-	// a pawn move must not specify the piece type
-	if (*it == 'x' || pieceFromSymbol(*it) == Pawn)
+	if (*it == 'x')
 		return Move();
-
+	// a pawn move should not specify the piece type
+	if (pieceFromSymbol(*it) == Pawn)
+		it++; // ignore character
 	// Piece type
 	Piece piece = pieceFromSymbol(*it);
 	if (piece.side() != Side::White)
@@ -496,6 +514,17 @@ QString WesternBoard::castlingRightsString(FenNotation notation) const
 	return str;
 }
 
+int WesternBoard::pawnAmbiguity(StepType t) const
+{
+	int count = 0;
+	for (const auto a: m_pawnSteps)
+	{
+		if (a.type == t)
+			++count;
+	}
+	return count;
+}
+
 QString WesternBoard::vFenString(FenNotation notation) const
 {
 	// Castling rights
@@ -504,6 +533,7 @@ QString WesternBoard::vFenString(FenNotation notation) const
 	// En-passant square
 	if (m_enpassantSquare != 0)
 		fen += squareString(m_enpassantSquare);
+		//TODO:  HOW to achieve Berolina disambiguation?
 	else
 		fen += '-';
 
@@ -631,17 +661,32 @@ bool WesternBoard::vSetFenString(const QStringList& fen)
 	m_sign = (side == Side::White) ? 1 : -1;
 	if (*token != "-")
 	{
-		setEnpassantSquare(squareIndex(*token));
+		int epSq = squareIndex(*token);
+		setEnpassantSquare(epSq);
 		if (m_enpassantSquare == 0)
 			return false;
 
+		Piece ownPawn(side, Pawn);
+		Piece opPawn(side.opposite(), Pawn);
+		int matchesOwn = 0;
+		int epTgt = 0;
+
+		for (const PawnStep& pStep: m_pawnSteps)
+		{
+			int sq = epSq - pawnPushOffset(pStep, m_sign);
+			Piece piece = pieceAt(sq);
+			if (pStep.type == CaptureStep && piece == ownPawn)
+				matchesOwn++;
+			else if (pStep.type == FreeStep && piece == opPawn)
+				epTgt = epTgt ? sq : sq;
+		}
 		// Ignore the en-passant square if an en-passant
 		// capture isn't possible.
-		int pawnSq = m_enpassantSquare + m_arwidth * m_sign;
-		Piece ownPawn(side, Pawn);
-		if (pieceAt(pawnSq - 1) != ownPawn
-		&&  pieceAt(pawnSq + 1) != ownPawn)
+		if (!matchesOwn)
 			setEnpassantSquare(0);
+		else
+			// set ep square and target
+			setEnpassantSquare(epSq, epTgt);
 	}
 
 	// Reversible halfmove count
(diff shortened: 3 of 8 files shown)


projects/lib/src/board/westernboard.h
```diff
@@ -76,6 +76,29 @@ class LIB_EXPORT WesternBoard : public Board
...
+		/*!
+		 * Movement mask for Pawn moves.
+		 * Forward: Left diagonal, straight, right diagonal.
+		 * \sa BerolinaBoard
+		 *
+		 * Default: A pawn can step straight ahead onto a free square or
+		 * capture diagonally forward.
+		 */
+		struct PawnStep { StepType type; int file; };
+		QList<PawnStep>  pawnSteps = { {Capture, -1}, {Free, 0}, {Capture, 1} };
```
ilaripih: We like to initialize member variables in the constructor instead of the declaration.

projects/lib/src/board/westernboard.h
```diff
@@ -163,17 +187,22 @@ class LIB_EXPORT WesternBoard : public Board
...
 		bool parseCastlingRights(QChar c);
 		CastlingSide castlingSide(const Move& move) const;
-		void setEnpassantSquare(int square);
+		void setEnpassantSquare(int square, int target=0);
 		void setCastlingSquare(Side side,
 				       CastlingSide cside,
 				       int square);
+		/*! Helper for Pawn moves. Returns square offset for the
+		 *  given \a step with orientation \a sign. */
+		inline int getPawnPush( const PawnStep& ps, int sign ) const;
```
ilaripih: Our convention doesn't use the "get" prefix unless necessary. I think a better name for this function would be `pawnPushOffset`.

projects/lib/src/board/berolinaboard.cpp
```diff
@@ -0,0 +1,44 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "berolinaboard.h"
+#include "westernzobrist.h"
```
ilaripih: It does not look like "westernzobrist.h" or "gaviotatablebase.h" are needed here.

ilaripih: It does not look like "westernzobrist.h" or "gaviotatablebase.h" are needed here.

alwey: Yes. gaviotatablebases.h is definitively out, westernzobrist.h too but could possible be needed later, but not for now.

projects/lib/src/board/berolinaboard.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace Chess {
+
+BerolinaBoard::BerolinaBoard()
+	: StandardBoard()
+{
+ /*
+  * Berolina pawns move one square diagonally forward and they capture
+  * one square straight ahead.
+  */
+	pawnSteps = { { Free, -1 }, { Capture, 0 }, { Free, 1 } };
```
alwey: This is no private class member. It is a protected member of WesternBoard. Being unsure I did not use m_ .

projects/lib/src/board/board.cpp
```diff
@@ -319,14 +319,15 @@ QString Board::moveString(const Move& move, MoveNotation notation)
 	return lanMoveString(move);
 }
 
-Move Board::moveFromLanString(const QString& str)
+Move Board::moveFromLanString(const QString& istr)
 {
+	QString str(istr);
+	str.remove(QRegExp("[x+#!?]"));
```
alwey: Of course you reacted to this. This touches a topic that needs some clarification for various reasons.

A more lenient reader is somewhat slower but may have advantages. Actually this is a *work-around* as long as the Berolina SAN notation is not agreed upon. The cutechess SAN reader does not like the longer pawn notation e2c4 and c5xc4  (which is Xboard-like) instead of ec4 and cxc4. So in the Board::moveFromString it will not get a valid moveFromSanString result and thus fall back and try moveFromLanString where a capture notation c5xc4 will spoil the result. To be able to read Xboard-scored Berolina games I made this change for now.

To resolve this I suggest to clarify two topics:

1. Berolina notation standards. I would prefer to have a common basis with authors of relevant software how to handle this.
2. Notation reader standards of cutechess. On the one hand the written and exported notations have to be strict and correct, but what is the goal of the reading methods? Should they be lenient and able to read from various sources or are they designed to be very strict and cough on small deviations?

Regarding the work-around, a correct SAN parser would find out about ambiguous source fields that result in a pawn move to the parsed target field. I have left that work out because I did not want to put effort into this before getting a  response from you.

projects/lib/src/board/westernboard.cpp
```diff
@@ -94,6 +95,13 @@ void WesternBoard::vInitialize()
 	m_rookOffsets[1] = -1;
 	m_rookOffsets[2] = 1;
 	m_rookOffsets[3] = m_arwidth;
+
+	m_pawnAmbiguous = pawnAmbiguity(StepType::Free);
+}
+
+inline int WesternBoard::getPawnPush(const PawnStep& ps, int sign) const
```
alwey: I do not know actually. I hope this is not a rhetoric question. I will investigate this. If I am not sure it works I will move the function body to the header file.

projects/lib/src/board/westernboard.cpp
```diff
@@ -165,6 +173,11 @@ QString WesternBoard::sanMoveString(const Move& move)
 
 	if (piece.type() == Pawn)
 	{
+		if ( m_pawnAmbiguous )
```
alwey: OK, I suspected this. Will be corrected. Of course a coding style with spaces is more readable at times.

projects/lib/src/board/berolinaboard.cpp
ilaripih: Oh right. I think only have one or two instances of protected member variables because they're generally a bad practice. But I think we still use the m_ prefix there.

projects/lib/src/board/westernboard.cpp
```diff
@@ -496,6 +510,15 @@ QString WesternBoard::castlingRightsString(FenNotation notation) const
 	return str;
 }
 
+bool WesternBoard::pawnAmbiguity( StepType t ) const
+{
+	int count = 0;
+	for (auto a: pawnSteps)
```
alwey: ACK. Corrected. 

--- Remark
Actually the method name is not good: pawnAmbiguity is a name for a quantifying method.
I thought about various names isBla... or hasBla... or pawnHasBla... to indicate a boolean result. I finally changed the method. It will now give the number of possible steps of the given type. So the method vInitialize now tests whether there are more than one FreeSteps.

projects/lib/src/board/westernboard.cpp
```diff
@@ -631,17 +655,32 @@ bool WesternBoard::vSetFenString(const QStringList& fen)
...
+		setEnpassantSquare(epSq);
 		if (m_enpassantSquare == 0)
 			return false;
 
+		Piece ownPawn(side, Pawn);
+		Piece opPawn(side.opposite(), Pawn);
+		int matchesOwn = 0;
+		int epTgt = 0;
+
+		for ( PawnStep pStep: pawnSteps )
```
alwey: I would like to avoid deep copies as long it is possible and makes sense. I am working on that.

projects/lib/src/board/westernboard.cpp
```diff
@@ -631,17 +655,32 @@ bool WesternBoard::vSetFenString(const QStringList& fen)
...
+		{
+			int sq = epSq - getPawnPush( pStep, m_sign );
+			Piece piece = pieceAt(sq);
+			if (pStep.type == Capture && piece == ownPawn)
+					matchesOwn++;
```
alwey: corrected

projects/lib/src/board/westernboard.cpp
```diff
@@ -771,20 +813,24 @@ void WesternBoard::vMakeMove(const Move& move, BoardTransition* transition)
...
 		// opportunity for the opponent.
-		else if ((source - target) * m_sign == m_arwidth * 2)
+		else if ((source / m_arwidth - target / m_arwidth) * m_sign == 2)
 		{
+			int epSq = (source + target) / 2;
 			Piece opPawn(side.opposite(), Pawn);
-			if (pieceAt(target - 1) == opPawn
-			||  pieceAt(target + 1) == opPawn)
-				setEnpassantSquare(source - m_arwidth * m_sign);
+			for ( PawnStep pstep: pawnSteps )
```
alwey: Ack.

projects/lib/src/board/westernboard.cpp
```diff
@@ -912,13 +958,15 @@ bool WesternBoard::inCheck(Side side, int square) const
...
-	int step = (side == Side::White) ? -m_arwidth : m_arwidth;
-	// Left side
-	if (pieceAt(square + step - 1) == Piece(opSide, Pawn))
-		return true;
-	// Right side
-	if (pieceAt(square + step + 1) == Piece(opSide, Pawn))
-		return true;
+	int sign = (side == Side::White) ? 1 : -1;
+
+	for ( PawnStep pStep: pawnSteps )
```
alwey: Ack.

projects/lib/src/board/westernboard.cpp
```diff
@@ -1051,41 +1099,33 @@ void WesternBoard::generatePawnMoves(int sourceSquare,
...
-			}
-		}
-	}
-
-	// Captures, including en-passant moves
+	// Normal moves, Captures, including en-passant moves
 	Side opSide(sideToMove().opposite());
-	for (int i = -1; i <= 1; i += 2)
+
+	for (PawnStep pStep: pawnSteps)
```
alwey: Ack.

projects/lib/src/board/westernboard.h
```diff
@@ -76,6 +76,29 @@ class LIB_EXPORT WesternBoard : public Board
 		/*! Movement mask for Rook moves. */
 		static const unsigned RookMovement = 8;
 
+		/*! Types of Pawn moves. */
+		enum StepType: const char
```
alwey: Good point - const char is a remnant of an idea I had in the start but I did not pursue. Deleted.
StepTypes: good idea - adopted.

projects/lib/src/board/westernboard.h
alwey: Well, I deliberately put it in the declaration part to have the default assignment close to the declaration of StepType for overview. I will change this.

projects/lib/src/board/westernboard.h
alwey: Yes, indeed a better name. This is no getter function. Renamed.

projects/lib/src/board/westernboard.cpp
```diff
@@ -1051,41 +1099,33 @@ void WesternBoard::generatePawnMoves(int sourceSquare,
...
-		targetSquare = sourceSquare - step + i;
+		targetSquare = sourceSquare + getPawnPush( pStep, m_sign );
 		capture = pieceAt(targetSquare);
-		if (capture.side() == opSide
-		||  targetSquare == m_enpassantSquare)
+		bool isCapture = capture.side() == opSide
+				||  targetSquare == enpassantSquare();
+		bool isNormalStep = capture.isEmpty();
+
+		if (  (isNormalStep && pStep.type == StepType::Free )
```
alwey: Ack-

projects/lib/src/board/westernboard.cpp
ilaripih: Yeah, the implicitly shared Qt containers haven't aged so well: https://www.kdab.com/goodbye-q_foreach/

projects/lib/src/board/westernboard.cpp
ilaripih: It's almost always possible to avoid deep copies, but with Qt containers not always obvious. The "Goodbye Q_FOREACH" blog post is a good guide on this. Because Qt containers use shared data, this does NOT result in a deep copy:

`const auto list = pawnSteps;`

But now you can safely iterate over `list` without an accidental deep copy, because the list is `const`.

projects/lib/src/board/berolinaboard.cpp
alwey: Adapted.

projects/lib/src/board/westernboard.cpp
alwey: https://isocpp.org/wiki/faq/inline-functions#inline-nonmember-fns

It should work as it is. The definition body would be needed in the header if more than one cpp-file referenced the function. Its a private member function for now. None of the other classes´ business.       

> "The reason you (almost always) put the definition (the {...} part) of an inline function in a header file is to avoid “unresolved external” errors from the linker. That error will occur if you put the inline function’s definition in a .cpp file and if that function is called from some other .cpp file."

alwey: I am sorry, I have not refreshed the page and so I did not  see your answer when I cited iscocpp.org.... I must remember to refresh the pages. Sometime I see automatic updates to discussions and sometimes not.

projects/lib/src/board/westernboard.h
```diff
@@ -76,6 +76,33 @@ class LIB_EXPORT WesternBoard : public Board
...
+		 * Movement mask for Pawn moves.
+		 * Lists pawn move types and relative files
+		 * Forward left diagonal -1, straight 0, right diagonal 1.
+		 * \sa BerolinaBoard
+		 *
+		 * Default: A Pawn can step straight ahead onto a free square or
+		 * capture diagonally forward. So initialise this
+		 * as { {CaptureStep, -1}, {FreeStep, 0}, {CaptureStep, 1} }
+		 */
+		QList<PawnStep>  m_pawnSteps;
```
ilaripih: The detaching problem still persists in the range-based for loops because the container `m_pawnSteps` is non-const. I suggest changing the container type to `QVarLengthArray<PawnStep>`. Then you don't have to change any of the for loops; it's enough to change this one line. We already use QVarLengthArray a lot in the Board classes because it's lean and efficient.

projects/lib/src/board/westernboard.h
ilaripih: The enum type is still `const char`.

alwey: oops, I removed it once. Probably forgot to check this change in. Thank you.

projects/lib/src/board/westernboard.h
alwey: Yes, sorry. That hasn't been enough to get rid of the local copy. A drawback of using `QVarLengthArray` is that one cannot use an initializer list to fill in the values in one step. I tried it before launching the first pull request.  Elements must be added individually. I just made a version using `QVarLengthArray` with changes in constructor bodies and initialiser. 

I am testing now and will upload the `QVarLengthArray` version soon.

I also have a version using a `const QList<PawnStep> &` as target of the range-based loop.

1.) rename m_pawnSteps to m_pawnStepsList)
`QList<PawnStep> m_pawnStepsList;`

2.) use m_pawnSteps identifier, now for a const QList reference)
`const QList<PawnStep>& m_pawnSteps;`

3.) assignments are done in the `WesternBoard::WesternBoard` init list.
`: Board(zobrist),
m_pawnStepsList {{CaptureStep, -1}, {FreeStep, 0}, {CaptureStep, 1}},
m_pawnSteps(m_pawnStepsList),
...`

As I understand, this would also work without copying the QList in the range-based loops?

ilaripih: Oh right, `QVarLengthArray` got initializer list support in Qt 5.5 but we need to support Qt 5.3. But that shouldn't be a major inconvenience in this case.

And yes, your suggestion would work.

alwey: Thank you. I stored a `const QList&` version away in my local repository. But the `QVarLengthArray` is used now in the HEAD version of the pull request. You might be more comfortable with this.

PR #174, opened 2016-10-14
do not use time increment for book moves

alwey:
Some users prefer to not increase available time if book moves are applied. Currently, any move causes a pre-defined time increment.  #9 

This simple patch does not have configurable settings.

### DIFF
--- a/projects/lib/src/timecontrol.cpp
+++ b/projects/lib/src/timecontrol.cpp
@@ -321,7 +321,7 @@ void TimeControl::startTimer()
 	m_time.start();
 }
 
-void TimeControl::update()
+void TimeControl::update(bool applyIncrement)
 {
 	/*
 	 * This will overflow after roughly 49 days however it's unlikely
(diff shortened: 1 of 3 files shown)


projects/lib/src/timecontrol.cpp
```diff
@@ -321,7 +321,7 @@ void TimeControl::startTimer()
 	m_time.start();
 }
 
-void TimeControl::update()
+void TimeControl::update(bool apply_increment)
```
ilaripih: Can you use our naming convention for variables (`m_memberVariable` for class members variables and `normalVariable` for others)? Then this parameter would be `applyIncrement`, `new_time_left` would be `newTimeLeft` etc.

alwey:
@ilaripih : /Wrt a recent talkchess.com discussion, please allow me to make a remark. I find, book moves should not be awarded with increment for various reasons, you already gave some of them in your considerations in #9.

> I'm also wondering if the book moves from the opening PGN are having the increments added to the total time? Lets say a 10 move depth opening book would add (10 x inc) time to the total time allocated to the engine. This is a problem if the book move depth is not uniform for all book entries. Some of the games will therefore have more time than the others depending on book move depth.

1. An opening sequence leads to a forced starting position. The conditions should be the same for a set-up by FEN or a set-up by PGN moves.

2. An opening book has non-uniform depths. Given an odd ply depth the white side is given an additional advantage of one increment and the black side has to move.

3. At a given time control with time increment the engines would be given the more time the deeper the book. This is really counter-intuitive.

4. As a consequence of point 3. testing conditions are more stable when no increments are rewarded to opening sequences.

ilaripih:
Agreed. The other position is almost purely ideological and we're all about being practical.

PR #175, opened 2016-10-14
begin support for variant King of the Hill

alwey:
This adds support for the variant kingofthehill ("King of the Hill")  to the library. It uses `StandardBoard` as base. So it is only necessary to apply the one extra rule and test for a win for a king on a center field.

It has been tested manually and using Sjaak II version 1.3.1a and a recent version of the Stockfish multi-variant fork of D. Dugovic and team https://github.com/ddugovic/Stockfish.git .

### DIFF
--- a/projects/lib/src/board/kingofthehillboard.h
+++ b/projects/lib/src/board/kingofthehillboard.h
@@ -1,0 +1,55 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


projects/lib/src/board/kingofthehillboard.h
```diff
@@ -0,0 +1,59 @@
...
+
+		// Inherited from StandardBoard
+		virtual Board* copy() const;
+		virtual QString variant() const;
+		virtual Result result();
+		/*!
+		 * Returns true if the king of \a side is occupying a central
+		 * square (one of d4, e4, d5, e5) else false.
+		 */
+		bool kingInCenter(Side side) const;
```
ilaripih: Can you make the `kingInCenter` function private?

alwey: I was already pondering about that. External code would probably not make use of `kingInCenter`. Changed and pushed.

Issue #148, opened 2016-09-17
Auto-flip board in engine (white) vs human (black) games


ilaripih:
In games where a human is playing against an engine as black, the board should be flipped, even though there's a separate "flip board" feature.

mcostalba:
And perhaps also in gauntlet mode to always follow from the first engine's side

ilaripih:
Done. Big thanks to @alwey !

PR #176, opened 2016-10-15
Automatically flip the board if human plays the black side against an engine

alwey:
It is more convenient to play with an automatically flipped board. #148

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -564,6 +564,10 @@ void MainWindow::setCurrentGame(const TabData& gameData)
 		m_evalWidgets[i]->setPlayer(player);
 	}
 
+	// if a human plays the black side against an engine then flip the board
+	if (m_players[Chess::Side::Black]->isHuman() && !m_players[Chess::Side::White]->isHuman())
+		m_gameViewer->boardScene()->flip();
+
 	updateWindowTitle();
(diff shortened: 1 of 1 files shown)


alwey:
please swap(the, plays) in title

projects/gui/src/mainwindow.cpp
```diff
@@ -564,6 +564,16 @@ void MainWindow::setCurrentGame(const TabData& gameData)
 		m_evalWidgets[i]->setPlayer(player);
 	}
 
+	// if a human plays the black side against an engine then flip the board
+	if (m_players[0]->isHuman() != m_players[1]->isHuman())
```
ilaripih: I don't think it makes sense to loop over the players when we know that the white player is always `m_players[0]` or `m_players[Chess::Side::White]` and black player is `m_players[1]` or `m_players[Chess::Side::Black]`. I think this should be all that is needed:

```
if (!m_players[Chess::Side::White]->isHuman()
&&  m_players[Chess::Side::Black]->isHuman())
	m_gameViewer->boardScene()->flip();
```

Also, as a side note, taking a reference to a pointer type (`const auto& player`) doesn't usually make sense.

alwey: Thank you. The simplified code is much better.

alwey:
Re-based and pushed new version.

PR #177, opened 2016-10-22
begin support for Racing Kings variant

alwey:
This adds support for the variant racingkings ("Racing Kings") to the library.  Racing Kings Chess is a race game and a checkless chess variant.

This software has been tested with a recent version of the multi-variant fork of the Stockfish engine  https://github.com/ddugovic/Stockfish.git. It also works with lichess.org PGN game files (after patching the variant name "Racing Kings" to "racingkings", but this is a topic already listed in the issues list).

### DIFF
--- a/projects/lib/src/board/racingkingsboard.cpp
+++ b/projects/lib/src/board/racingkingsboard.cpp
@@ -1,0 +1,128 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "racingkingsboard.h"
+#include "westernzobrist.h"
+
+namespace Chess {
+
+RacingKingsBoard::RacingKingsBoard()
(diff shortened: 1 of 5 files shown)


projects/lib/src/board/racingkingsboard.cpp
```diff
@@ -0,0 +1,132 @@
...
+	Square ksq = chessSquare(kingSquare(side));
+	return (ksq.rank() == 7);
+}
+
+/*! Returns true if the king of \a side can reach the eighth rank */
+bool RacingKingsBoard::canFinish(Side side)
+{
+	QVarLengthArray<Move> moves;
+	WesternBoard::generateMovesForPiece(moves, King, kingSquare(side));
+	for (const Move &m: moves)
```
ilaripih: One coding convention nit-pick: we attach the reference/deference operator to the type; not the variable name. So the for loop should be `for (const Move& m : moves)`.

projects/lib/src/board/racingkingsboard.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+	return "racingkings";
+}
+
+QString RacingKingsBoard::defaultFenString() const
+{
+	return "8/8/8/8/8/8/krbnNBRK/qrbnNBRQ w - - 0 1";
+}
+
+bool RacingKingsBoard::vIsLegalMove(const Move& move)
```
ilaripih: It doesn't look like you need to reimplement `vIsLegalMove`. `Board::vIsLegalMove` already looks like this:

```
	makeMove(move);
	bool isLegal = isLegalPosition();
	undoMove();

	return isLegal;
```

So you could just reimplement `WesternBoard::isLegalPosition` like this:

```
bool RacingKingsBoard::isLegalPosition()
{
	if (inCheck(sideToMove()))
		return false;
	return WesternBoard::isLegalPosition();
}
```

I didn't test it so it's possible I missed something.

projects/lib/src/board/racingkingsboard.cpp
alwey: Changed, I will upload a new version soon.

Issue #181, opened 2016-11-16
cutechess-cli adjudication with TB doesn't take 50-move rule into account


Prcuvu:
When it reaches a TB position which is a draw under 50-move rule but a win without the rule, cutechess-cli always adjudicates to a win.

Edit: cutechess still uses Gaviota TB, seems more necessary to switch to syzygy.

Reference:
* http://tcec.chessdom.com/archive.php?se=9&rapid&ga=312
* http://tcec.chessdom.com/archive.php?se=9&sf&ga=17
* #68

ilaripih:
Syzygy solves this and there's a unit test for the TCEC position in question.

Issue #68, opened 2016-06-07
replace Gaviota tablebases with Syzygy


ilaripih:
Syzygy would be better because:
- Better compression
- 6-men tablebases
- Defects in the GTB code (out-of-bounds reads, resource leaks etc)
- Stockfish already has a good C++ probing implementation that we can use. And its a lot less code than GTB: https://github.com/official-stockfish/Stockfish/tree/master/src/syzygy

ilaripih:
The Stockfish Syzygy probe is actually being rewritten: https://github.com/official-stockfish/Stockfish/tree/syzygy

artoj:
Would it make sense to adapt this code to Qt / C++11 style? I see that it has code dedicated to Win32/Unix compatibility that could be thrown out.

artoj:
It seems that the new code uses C++11.

raimarHD:
Here is another approach to integrate Syzygy into an existing program.
https://github.com/basil00/Fathom
Some chessengines like Arasan use this lib.

ilaripih:
Thanks for letting us know! I like Stockfish's C++ code (closer to most CuteChess code) but Fathom would actually be significantly easier to integrate into CuteChess because we wouldn't have to rewrite any Syzygy code.

mcostalba:
@ilaripih In case you opt to use SF syzygy code, please use this file:

https://github.com/official-stockfish/Stockfish/blob/syzygy/src/syzygy/tbprobe.cpp

and delete all the lines starting from [here](https://github.com/official-stockfish/Stockfish/blob/syzygy/src/syzygy/tbprobe.cpp#L1477) til the EOF. This code has not been updated and is still in the original form, I have not touched it because I planned to completely drop it because hacky and useless.

In you case you just need:

```
void init(const std::string& paths);
WDLScore probe_wdl(Position& pos, ProbeState* result);
int probe_dtz(Position& pos, ProbeState* result);
```

header is [here ](https://github.com/official-stockfish/Stockfish/blob/syzygy/src/syzygy/tbprobe.h)

Probing code is fully tested in parallel with original code to ensure all the calls to both code paths  return same values (tested on many hundreds millions calls over hundreds of endgame positions), and actually does not have a bug that instead is present in the original code.

Finally I think the code properly documents how TB works.

mcostalba:
@ilaripih in case you are interested I can factor the needed code out of SF so to make a kind of library.

ilaripih:
[...]

PR #183, opened 2016-11-20
Replace the Gaviota tablebases with Syzygy.

basil00:
This allows for:
    - Up to 6 piece tablebase adjudication.
    - Proper handling of the 50 move rule.

The implementation uses https://github.com/basil00/Fathom which is included as a "3rd party library".

See also #181, #68 and [the TCEC9 superfinal game 17 controversy](http://talkchess.com/forum/viewtopic.php?t=62145).


ilaripih:
Wow, that's awesome! I've been planning on doing this for a while but have been too busy lately. I'll do a code review of the CuteChess/C++ parts (not Fathom) on Monday. If you check the CI reports you'll see that the GCC builds are currently failing: https://travis-ci.org/cutechess/cutechess/jobs/177367488
Looks like the problem can be fixed by compiling Fathom with `-std=c99`.

ilaripih:
I also noticed this Fathom issue: https://github.com/basil00/Fathom/issues/13

Cute Chess runs each game in its own thread so it can make multiple `tb_probe_root` calls in parallel. Those calls should be protected by a `QMutex`.

basil00:
> Those calls should be protected by a QMutex.

Yes you are right -- I'll fix this and the gcc issue when I have some more free time.

Another (currently unimplemented) idea is to make TB adjudication depend on whether the engines themselves support TBs.  Quoting [Ronald de Man](http://talkchess.com/forum/viewtopic.php?t=62145&postorder=asc&start=170):

`I personally don't think 6-piece TB adjudication is a good idea if one of the engines does not use TBs. There are many 6-piece positions that (some) engines are not able to play correctly, so adjudicating those is helping TB-less engines too much. This problem also exists for 5-piece positions, but to a much smaller extent.`

The question would be how to detect an engine's TB configuration, e.g. via UCI options?

ilaripih:
I think the operator/user just has to know what TBs the engines can use and set the adjudication rules accordingly. It is possible to detect the engines' tablebase settings but they're not fully standardized.

ilaripih:
I reviewed the rest of the code (excluding the 3rd party stuff) and it looks good. So the mutex and compilation flag are the only issues.

basil00:
> I think the operator/user just has to know what TBs the engines can use and set the adjudication rules accordingly. It is possible to detect the engines' tablebase settings but they're not fully standardized.

Yes - I think it might be worthwhile to add an extra engine option that can be used to disable TB adjudication for specific engines, e.g. something like:

    tbadj=[on | off]

The default is `on` provided TBs are available.  I might work on a proposal as a separate PR.

basil00:
> I reviewed the rest of the code (excluding the 3rd party stuff) and it looks good. So the mutex and compilation flag are the only issues.

Appears to now be fixed.

PR #185, opened 2016-12-03
support Horde Chess

alwey:
This is a proposal to support the Horde Chess variant(s). Horde chess is a variant that is similar to Dunsany's chess of 1942. A standard chess set of pieces faces a horde of pawns. The pieces must eliminate all pawns to win. The pawns go to checkmate the opponent king. Standard rules apply,

This patch implements the modern variant with 36 white pawns (lichess.org, stockfish). It is quite general and can also be used with switched black/white setup (like older Horde variant). 

The version was tested using stockfish variant fork https://github.com/ddugovic/Stockfish.
I had to adapt the en passant rules to not accept e.p. on rank 2 but to allow double steps from rank 1 (matching perft of stockfish).

A change to WesternBoard makes the assertion of number of kings more flexible (the horde has no king).

I hope this is useful.

### DIFF
--- a/projects/lib/src/board/board.pri
+++ b/projects/lib/src/board/board.pri
@@ -7,6 +7,7 @@ SOURCES += $$PWD/board.cpp \
     $$PWD/racingkingsboard.cpp \
     $$PWD/capablancaboard.cpp \
     $$PWD/kingofthehillboard.cpp \
+    $$PWD/hordeboard.cpp \
     $$PWD/zobrist.cpp \
     $$PWD/westernzobrist.cpp \
     $$PWD/frcboard.cpp \
@@ -31,6 +32,7 @@ HEADERS += $$PWD/board.h \
     $$PWD/racingkingsboard.h \
     $$PWD/capablancaboard.h \
     $$PWD/kingofthehillboard.h \
+    $$PWD/hordeboard.h \
     $$PWD/zobrist.h \
     $$PWD/westernzobrist.h \
     $$PWD/frcboard.h \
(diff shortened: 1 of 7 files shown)


ilaripih:
Thanks! At a first glance the code looks good, but unfortunately there's some instability in move generation and/or making/unmaking moves, highlighted by these two failing Perft tests:

* https://travis-ci.org/cutechess/cutechess/jobs/180976730
* https://travis-ci.org/cutechess/cutechess/jobs/180976723

The "horde dunsany startpos" count is wrong in both runs, but interestingly the result is not the same. The perft tests are multithreaded so I suspect the cause is one of the following:

* Race condition
* Buffer overflow or some counter overflows due to the unexpectedly high number of pieces on the board.

ilaripih:
I think I solved it - `WesternBoard::inCheck` could not handle `m_kingSquare[side]` being 0 for the kingless side. The easiest fix is to modify the beginning of `WesternBoard::inCheck` to look like this:

```
	Side opSide = side.opposite();
	if (square == 0)
	{
		square = m_kingSquare[side];
		// In the "horde" variant the horde side has no king
		if (square == 0)
			return false;
	}
```

Can you please add this fix to your pull request?

alwey:
Great job! I tested this over and over. No problems on my computer. I was alarmed to see the check fail. But hey - good to have these multithreaded tests. Thank you. I will work your code in.

alwey:
Some problems with Qt 5.3.1 build. Horde perft tests are ok.

ilaripih:
It's all good. The Qt 5.3 builds failed because the third-party Qt repository couldn't be added for some reason. These are usually temporary problems.

PR #186, opened 2016-12-03
support for Three-Check Chess

alwey:
This patch adds support for Three-Check (actually N-Check) Chess.
Giving Check n times is an additional way to win a chess game.

Variant `3check` was tested manually and also using PyChess, Sjaak II and the Stockfish fork (https://github.com/ddugovic/Stockfish). For my tests I patched stockfish for that variant string, it normally uses `threecheck`).

This version can read various FEN formats, including standard chess FEN and extended FENs from lichess.org with check counters appended to the FEN (like <fen> +0+0) as well as a format suggested by H. G. Muller and implemented into Stockfish. The latter uses backward counters like <fen up to e. p. field> 3+3 <rest of fen string> and only one plus sign.  The writing format yet has to be determined. Currently reverse counters are appended to the fen: <fen> 3+3.

I hope this is useful.

### DIFF
--- a/projects/lib/src/board/ncheckboard.cpp
+++ b/projects/lib/src/board/ncheckboard.cpp
@@ -1,0 +1,193 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "ncheckboard.h"
+
+
+namespace Chess {
+
+/*! Maximum feasible value \a maxNCheck for nth-check limit. */
+static const int maxNCheck = 99;
+
+NCheckBoard::NCheckBoard(int n)
+	: StandardBoard(),
+	  m_checkLimit(qBound(1, n, maxNCheck))
+{
+}
+
+Board* NCheckBoard::copy() const
+{
+	return new NCheckBoard(*this);
+}
+
+QString NCheckBoard::variant() const
+{
+	return QString("%1check").arg(checkLimit());
+}
+
+QString NCheckBoard::defaultFenString() const
+{
+	return QString("rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - %1+%1 0 1")
+	.arg(checkLimit());
+}
+
+void NCheckBoard::vInitialize()
+{
+	m_checksToWin[Side::White] = m_checksToWin[Side::Black] = checkLimit();
+	StandardBoard::vInitialize();
+}
+
+void NCheckBoard::updateCheckCounters(Side side, int d = -1)
+{
+	Q_ASSERT(m_checksToWin[Side::White] >=0);
+	Q_ASSERT(m_checksToWin[Side::Black] >=0);
+	Q_ASSERT(side != Side::NoSide);
+
+	if (inCheck(side.opposite()))
+		m_checksToWin[side] += d;
+}
+
+void NCheckBoard::vMakeMove(const Move& move,
+			    BoardTransition* transition)
+{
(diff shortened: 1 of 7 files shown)


projects/lib/src/board/ncheckboard.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+void NCheckBoard::vInitialize()
+{
+	m_checksToWin[Side::White] = m_checksToWin[Side::Black] = checkLimit();
+	StandardBoard::vInitialize();
+}
+
+void NCheckBoard::updateCheckCounters(Side side, int d = -1)
+{
+	Q_ASSERT (m_checksToWin[Side::White] >=0);
```
ilaripih: Our convention is to use Q_ASSERT without a space before the parenthesis, and to add a space before and after operators (eg. `>=`):

`Q_ASSERT(m_checksToWin[Side::White] >= 0);`

projects/lib/src/board/ncheckboard.cpp
```diff
@@ -0,0 +1,197 @@
...
+	Q_ASSERT (m_checksToWin[Side::Black] >=0);
+	Q_ASSERT (side != Side::NoSide);
+
+	if ( inCheck(side.opposite()) )
```
ilaripih: Our convention is also not to pad parenthesis with spaces so this should be:

`if (inCheck(side.opposite()))`

projects/lib/src/board/ncheckboard.cpp
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * Our method however *appends* such a check counter field to the FEN.
+ * It is more compatible with standard chess data and software.
+ *
+ * In contrast Lichess uses two plus signs and *forward counters* (like +0+0)
+ * appended to the normal FEN.
+ */
+QString NCheckBoard::vFenString(Board::FenNotation notation) const
+{
+	QString fen = StandardBoard::vFenString(notation);
```
ilaripih: I would also use the `QString::arg` convention here:

```
QString fen = QString("%1 %2+%3")
	.arg(m_checksToWin[Side::White])
	.arg(m_checksToWin[Side::Black]);
```

projects/lib/src/board/ncheckboard.cpp
```diff
@@ -0,0 +1,197 @@
...
+bool NCheckBoard::vSetFenString(const QStringList& fen)
+{
+	if (fen.count() < 2)
+		return false;
+
+	// Make a copy to modify and to forward to parent's method
+	QStringList sfen(fen);
+
+	// Reset counters, do not rely on getting check counters from PGN
+        setChecksToWin(checkLimit(), checkLimit());
```
ilaripih: Indentation fail ;-)

alwey:
Thank you! I will work it over accordingly.

alwey:
Some discussion with @HGMuller ongoing /wrt three-check FEN extension. I am open minded.

ilaripih:
Thanks for the fixes and for pointing out the ongoing discussion. As I mentioned there, I think @HGMuller is right about the extra FEN tokens - they should appear before the ply counter and move number tokens.

alwey:
We should adapt the FEN writer accordingly. The reader has no problems. It also is important to investigate the communication and setup of engines. I can do that for stockfish, PyChess and Sjaak II.

Board and WesternBoard already have a good FEN structure.  Perhaps vFenString can be amended.

alwey:
Now FENs are in suggested format.

projects/lib/src/board/ncheckboard.cpp
```diff
@@ -0,0 +1,192 @@
...
+}
+
+QString NCheckBoard::variant() const
+{
+	return QString("%1check").arg(checkLimit());
+}
+
+QString NCheckBoard::defaultFenString() const
+{
+	return QString("%1 %2+%2").arg(StandardBoard::defaultFenString()).arg(checkLimit());
```
ilaripih: This FEN is still in the old format. I think it's easiest just to use this:

`QString("rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq -%1 0 1").arg(checkLimit())`

alwey: Actually we do not need NCheckboard::defaultFenString() any more.

ilaripih: Sorry I don't understand, can you elaborate why it's not needed?

alwey: Thank you, this slipped by, sorry. I changed the code accordingly.

alwey: The code worked without defaultFenString() because defaults of StandardBoard and NCheckBoard came into action but it is no clean solution. I probably have overlooked something anyway.

PR #189, opened 2016-12-11
flip board for gauntlet player 0 with black pieces

alwey:
Following a gauntlet is easier if seen from the gauntlet player's side (here engine with index 0).
As requested in #148 .

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -565,7 +565,14 @@ void MainWindow::setCurrentGame(const TabData& gameData)
 	}
 
 	// if a human plays the black side against an engine then flip the board
-	if (m_players[Chess::Side::Black]->isHuman() && !m_players[Chess::Side::White]->isHuman())
+	// also flip the board if gauntlet player 0 has black pieces
+	const Tournament* t = m_game->tournament();
+	bool gPlayer0Black = t
+			  && t->type() == "gauntlet"
+			  && t->playerIndex(m_game, Chess::Side::Black) == 0;
+
+	if (!m_players[Chess::Side::White]->isHuman()
+	&& ( m_players[Chess::Side::Black]->isHuman() || gPlayer0Black ))
 		m_gameViewer->boardScene()->flip();
 
 	updateWindowTitle();
--- a/projects/lib/src/chessgame.h
+++ b/projects/lib/src/chessgame.h
@@ -28,6 +28,7 @@
 #include "board/move.h"
 #include "timecontrol.h"
 #include "gameadjudicator.h"
+#include "tournament.h"
 
 namespace Chess { class Board; }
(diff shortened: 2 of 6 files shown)


alwey:
The Game Wall still has to be addressed.

projects/gui/src/mainwindow.cpp
```diff
@@ -565,7 +565,15 @@ void MainWindow::setCurrentGame(const TabData& gameData)
...
 	// if a human plays the black side against an engine then flip the board
-	if (m_players[Chess::Side::Black]->isHuman() && !m_players[Chess::Side::White]->isHuman())
+	if (m_players[Chess::Side::Black]->isHuman()
+	&& !m_players[Chess::Side::White]->isHuman())
+		m_gameViewer->boardScene()->flip();
+
+	// flip the board if gauntlet player 0 has black pieces
+	else if (gameData.tournament
+	&&  gameData.tournament->type() == "gauntlet"
+	&&  gameData.tournament->playerAt(0).name() == m_players[Chess::Side::Black]->name()
```
ilaripih: I'd prefer to compare player indexes instead of names if possible, eg:

```
// Check if the black player is the gauntlet player
if (gameData.tournament->m_gameData[m_game]->blackIndex == 0)
```

Of course `Tournament::m_gameData` is private so there needs to be a new helper method, eg. `int Tournament::playerIndex(ChessGame* game, Chess::Side side) const`.

alwey:
The GameWall now also flips boards.

projects/lib/src/chessgame.h
```diff
@@ -137,6 +142,7 @@ class LIB_EXPORT ChessGame : public QObject
 		QSemaphore m_pauseSem;
 		QSemaphore m_resumeSem;
 		GameAdjudicator m_adjudicator;
+		const Tournament* m_tournament;
```
ilaripih: I'll have to think about this a little. I'd like to avoid the two-way dependency between `ChessGame` and `Tournament`.

alwey: The `GameWall` only knows individual games. So I reluctantly added `m_Tournament` to `ChessGame`. I originally thought of a `signal`/`slot` solution to notify `GameWall`.

ilaripih: I can see the logic there. Anyway, I thought I'd put code behind my vision so I hope you don't mind that I rewrote most of your code. It's still not as pretty as I would like but achieves a couple of goals:

* Avoid copy-pasting the same board flipping logic to multiple places
* Make sure `ChessGame` is not dependent on `Tournament`
* Put `GauntletTournament` specific code inside the `GauntletTournament` class, which is cleaner than checking the tournament type from outside.

ilaripih:
A modified version of the commit is now in master -> closing the PR

PR #194, opened 2017-01-06
Berolina FEN ep extension

alwey:
This patch adds a FEN extension for variants with ambiguous en-passant cases.
The en-passant field comprises of ep square (destination square of capture move) and ep target (square of capture victim) like `d6e5`. 
This format was introduced by H. G. Muller and implemented into xboard (development version > 4.9.1).


alwey:
Note: Evert Glebbeek already implemented this extension into Sjaak II before version 1.0.0 (of 2015).
This works well with sjaakii.

Issue #157, opened 2016-09-26
Confirmation for closing active tournament game


elcabesa:
when playing a tournament with more than one concurrent game you can follow a game selecting it from the upper tab.
if you accidentally press the red X in the tab you close the tab and the game ends with a disconnection.
I thinnk this is unwanted, maybe you can disable it for tournament or popup a confirmation window. ( possibile without stopping the tournament)

thank you

ilaripih:
I still think it should be possible to force an end to a tournament game, but there should be a confirmation dialog where the user could choose a result for the game or cancel the stoppage.

ilaripih:
There's now a confirmation dialog. Choosing a result for the game may come later if it's needed.

PR #196, opened 2017-01-08
only protect a tournament game tab if the corresponding game is active

alwey:
At the end of a tournament tab message box and results window have a tendency to get into each other's way.

So I suggest to only protect a tourney game tab if the corresponding game is active. I made some malicious tests already (for running games, close while waiting between games, before and after tournament finished, also with different delays of confirmation). #157


ilaripih:
Oh of course, thanks!

PR #197, opened 2017-01-20
support Loop Chess and Chessgi

alwey:
This is a proposal to support the _Loop Chess_ and _Chessgi_ variants. Loop Chess is a variant that is similar to _Crazyhouse_. Captured pieces change sides and can be dropped onto empty squares later. In contrast to Crazyhouse, promoted pawns keep their piece rank when captured in Loop Chess. 

The implementation uses `CrazyhouseBoard`, which is modified to allow overriding promoted piece types. I tested with recent Stockfish multivariant versions and with TJchess 1.3.

Chessgi differs from Loop Chess only in drop rule for pawns: Chessgi allows to drop pawns on a side's own first rank.

The implementation of `ChessgiBoard` relies on `LoopBoard`. Tests were done with Sjaak II versions 1.3.1a and 1.4.1. Currently cutechess allows double steps from first rank and second rank, Sjaak II only from second rank (perhaps this can be made selectable in WesternBoard in the future).

I hope this is useful.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -47,8 +47,10 @@ Berolina Chess
 Capablanca Chess
 .It caparandom
 Capablanca Random Chess
+.It chessgi
+Chessgi / Drop Chess
 .It crazyhouse
-Crazyhouse / Drop Chess
+Crazyhouse (Drop Chess Variant)
 .It fischerandom
 Fischer Random Chess / Chess 960
 .It gothic
@@ -57,6 +59,8 @@ Gothic Chess
 Horde Chess (v2)
 .It kingofthehill
 King of the Hill Chess
+.It loop
+Loop Chess (Drop Chess Variant)
 .It losers
 Loser's Chess
 .It racingkings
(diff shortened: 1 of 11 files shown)


ddugovic:
"Currently cutechess allows double steps from first rank and second rank"

I have asked lichess regarding Horde chess regarding their 1st-rank double-push en passant rule & it was a struggle for me to implement that in Stockfish.  I prefer @alwey 's easier to understand rules (assuming en passant remains legal after any double pawn-push).

ilaripih:
Great work, thanks!

alwey:
@ilaripih @ddugovic: Thank you. I much prefer using double steps for ranks one and two. It has been implemented in cutechess like this,  and I find it logical. This detail had not become visible to the users  before _Chessgi_ and _Horde_.  Of cause I would adapt for variants with a clear and explicit tradition of rank-two-only doublesteps.

En-passant is normally coupled to double- or multi-steps. So there is consensus. 

@HGMuller: I know that you have pointed out the necessity of clarification for rank one pawns since long.

PR #199, opened 2017-01-22
support Extinction Chess and Kinglet Chess

alwey:
This is a proposal to support the _Extinction Chess_ and _Kinglet Chess_ variants.

In _Extinction Chess_ a side wins if their opponent loses all pieces of any kind. The King is a normal piece, there is no check and Pawns can also promote to King.

_Kinglet Chess_ is an older variant: A side wins if their opponent lost all Pawns. Pawns can promote to King only. 

I only tested manually. The engines I have either do not play these variants or are very old and require some more work on the interfaces. Nevertheless I hope this is useful.

### DIFF
--- a/projects/lib/src/board/extinctionboard.cpp
+++ b/projects/lib/src/board/extinctionboard.cpp
@@ -1,0 +1,146 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "extinctionboard.h"
+
+namespace Chess {
+
+ExtinctionBoard::ExtinctionBoard(const QSet<int>& set,
+				 bool allPromotions)
+	: StandardBoard(),
+	  m_allPromotions(allPromotions)
+{
+	for (const auto& s: set)
+		m_pieceSet.append(s);
+}
+
+Board* ExtinctionBoard::copy() const
+{
+	return new ExtinctionBoard(*this);
+}
+
+QString ExtinctionBoard::variant() const
+{
+	return "extinction";
+}
+
+QString ExtinctionBoard::defaultFenString() const
+{
+	return "rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1";
+}
+
+bool ExtinctionBoard::kingsCountAssertion(int, int) const
+{
+	return extinctPiece(Side::White).isEmpty()
+	||     extinctPiece(Side::Black).isEmpty();
+}
+
+void ExtinctionBoard::addPromotions(int sourceSquare,
+				int targetSquare,
+				QVarLengthArray<Move>& moves) const
+{
+	if (m_allPromotions)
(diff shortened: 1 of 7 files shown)


projects/lib/src/board/extinctionboard.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+	return false;
+}
+
+Piece ExtinctionBoard::extinctPiece(Side side) const
+{
+	for (Piece pc: m_pieceSet)
+	{
+		int cnt = 0;
+		for (int i = 0; cnt < 1 && i < arraySize(); i++)
```
ilaripih: This is clever, but it mislead me a bit because I thought we were counting things when we actually just wanted to find out if there's at least one piece on the board. I think it would be cleaner with a boolean variable and also faster because we could avoid the `cnt < 1` comparison at every step:

```
bool pieceFound = false;
for (int i = 0; i < arraySize(); i++)
{
	Piece piece = pieceAt(i);
	if (side == piece.side() && pc.type() == piece.type())
	{
		pieceFound = true;
		break;
	}
}
if (!pieceFound)
	return Piece(side, pc.type());
```

projects/lib/src/board/extinctionboard.h
```diff
@@ -0,0 +1,101 @@
...
+		virtual bool kingsCountAssertion(int whiteKings,
+						 int blackKings) const;
+		virtual bool inCheck(Side side, int square = 0) const;
+		virtual void addPromotions(int sourceSquare,
+					   int targetSquare,
+					   QVarLengthArray<Move>& moves) const;
+
+	private:
+		bool m_allPromotions;
+		QVarLengthArray<Piece> m_pieceSet;
```
ilaripih: I think this should be `QVarLengthArray<int> m_pieceTypes` instead because the pieces' side is never used, just the type. And `Piece::type()` is and `int`.

Then it would be neat if the `set` parameter to the constructor could be an initializer list, but unfortunately that's not supported by Qt versions < 5.5. So there would probably have to be a `setPieceTypes` method that `KingletBoard` could call, or you could have a virtual method returning the piece types for `extinctPiece()`.

projects/lib/src/board/extinctionboard.h
```diff
@@ -0,0 +1,101 @@
...
+ * The king has no royal obligation, so there is no check and the king can
+ * be captured. Castling is allowed. Stalemate is a draw.
+ *
+ * \note Rules: https://en.wikipedia.org/wiki/V._R._Parton#Kinglet_Chess
+ * \sa Extinction Chess
+ */
+class LIB_EXPORT KingletBoard : public ExtinctionBoard
+{
+	public:
+		KingletBoard(): ExtinctionBoard("P", false){}
```
ilaripih: Even though these member functions are one-liners, can you move the implementation to the .cpp file for consistency with the rest of the codebase?

projects/lib/src/board/extinctionboard.cpp
```diff
@@ -0,0 +1,122 @@
...
+			return Piece(side, pc.type());
+	}
+	return Piece();
+}
+
+Result ExtinctionBoard::result()
+{
+	QString str;
+	Piece piece = extinctPiece(sideToMove());
+	if (piece != Piece::NoPiece)
```
ilaripih: We have the `Piece::isValid()` method for this. In this case `!Piece::isEmpty()` would be okay as well.

projects/lib/src/board/extinctionboard.cpp
alwey: Thank you. True, I should have reworked this loop for clarity. Changed.

projects/lib/src/board/extinctionboard.cpp
alwey: Changed

projects/lib/src/board/extinctionboard.h
alwey: Thank you, using integer is better.

projects/lib/src/board/extinctionboard.h
alwey: Ok, clear coding style.

alwey:
Updated after code review: eliminated a problem /wrt to promotions and added another test case for _Kinglet_.

PR #200, opened 2017-01-24
support Losing Chess (Antichess, Giveaway, Suicide)

alwey:
This is a proposal to support  _Losing Chess_  variants _Antichess_, _Giveaway_, _Suicide_. (Currently there are some problems.)

In _Losing Chess_ a side that lost all pieces or cannot move wins. The King is a normal piece, there is no check and Pawns can also promote to King. If a side can capture a piece, the must do so.

The variant labelled _Antichess_ is the oldest and has no castling moves. _Giveaway_ is identical, but castling is allowed. _Suicide Chess_ is like _Antichess_ but in stalemate the side with less pieces wins.

`GiveawayBoard` and `SuicideBoard` were implemented as subclasses of `AntiBoard`.

I tested _Giveaway_ and _Suicide Chess_ with recent multivariant Stockfish and Sjeng 11,2. The latter was also tested using Sjaak II 1.3.1a and PyChess. _Antichess_ was test manually and by imports from Lichess.org.

In contrast to other variants I observed stability problems with the UCI interface when Stockfish engines went to depth rapidly with very low effective branching. I noticed move data corruption during moveFromString execution (Issue #201) under heavy load from `info pv` line processing. This includes conversion to SAN and testing legality of PV moves.

For now, I made a workaround to skip SAN conversion and legality testing of PVs. This reduces failure rate drastically. This helped me to find some other, less frequent problems. My attempts to use a mutex on some execution paths the have not been successful so far.

I hope this is useful nevertheless.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -43,6 +43,8 @@ Five-check Chess
 Andernach Chess
 .It antiandernach
 Anti-Andernach Chess
+.It antichess
+Antichess / Losing Chess
 .It atomic
 Atomic Chess
 .It berolina
@@ -61,6 +63,8 @@ Crazyhouse (Drop Chess Variant)
 Extinction Chess
 .It fischerandom
 Fischer Random Chess / Chess 960
+.It giveaway
+Giveaway Chess (Losing Chess)
 .It gothic
 Gothic Chess
 .It horde
@@ -79,6 +83,8 @@ Loop Chess (Drop Chess Variant)
 Loser's Chess
 .It racingkings
 Racing Kings Chess
+.It suicide
+Suicide Chess (Losing Chess Variant)
 .It superandernach
 Super-Andernach Chess
 .It standard
--- a/projects/cli/res/doc/help.txt
+++ b/projects/cli/res/doc/help.txt
@@ -14,6 +14,7 @@ Options:
 			'5check': Five-check Chess
 			'andernach': Andernach Chess
 			'antiandernach': Anti-Andernach Chess
+			'antichess': Antichess (Losing Chess)
 			'atomic': Atomic Chess
 			'berolina': Berolina Chess
 			'capablanca': Capablanca Chess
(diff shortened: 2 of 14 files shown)


ilaripih:
Thanks for the new variants. Did you notice that Cute Chess already has `LosersBoard` which I think is the same as AntiChess: https://github.com/cutechess/cutechess/blob/master/projects/lib/src/board/losersboard.cpp

alwey:
Thank you. 

> Did you notice that Cute Chess already has LosersBoard which I think is the same as AntiChess: https://github.com/cutechess/cutechess/blob/master/projects/lib/src/board/losersboard.cpp


No LosersBoard is a different matter. Losers has a royal King, there is check and mate. In antichess you can lose the King or have multiple Kings. This is why I made an extension to `WesternBoard` to remove castling rights for Giveaway when the king is captured on the base rank.

ilaripih:
Oh, okay. I'll test the data corruption issue later. GCC's and LLVM's address sanitizer should help in detecting issues.

ianfab:
First of all, great work! Online testing for multi-variant Stockfish depends on cutechess support for variants, so it would be great to have support for losing chess variants. Therefore, I would like to know the current status of this PR. I have not done extensive testing, but in a few hundred giveaway chess games (TC 10+0.1) with current master of multi-variant Stockfish, I have not observed any errors with this PR. So do you think this is stable enough to be used for engine testing?

alwey:
@ianfab: Great to hear, this patch uses a workaround that skips SAN conversion for PV in _antichess_, _giveaway_ and _suicide_ variants. This is a hack but it might be OK for expert use, e.g. for _fishtest_. I think it is not good enough for a final version however. I observed that on my machine the workaround decreased the error rate by factor 1/200 for 30s+0.03s games. I am keen to hear more about this.

ianfab:
@alwey Thanks, I'll consider trying it on fishtest.

ianfab:
@alwey FYI: We now use this PR on [fishtest](http://35.161.250.236:6543/tests) and everything looks good so far on several different machines (linux/windows, concurrency from 1 up to 15) with currently about 40000 giveaway chess games in total.

alwey:
@ianfab: Great to see these tests on fishtest.
I hope the game files do not have many lines like: ` [Termination "illegal move"]`, because these would probably be a problem of the CLI rather than one of Stockfish.

alwey:
This version includes a work-around against problems with move data by using direct engine PV output without legality test. In this PR version the variant name is only tested at game start against a `const QList` of names instead of doing this for every PV line. For this a new flag `m_useDirectPv` is used.

Issue #198, opened 2017-01-21
FEN handling (read) of promoted pieces in Crazyhouse variant


alwey:
At setup, `CrazyhouseBoard` does not decode FEN correctly for promoted pieces, e.g.

`3q1bkr/2p1pBp1/q1n3p1/1N2p3/1Pp5/P4Q~2/BBPp1PPP/R2K2NR[RPPn] b - - 0 28` .

The parser does not like the mixed one-letter/two-letter piece descriptions, e.g. ` Q `and `Q~`
For `Q~` it will decode `Q` because the first character matches a valid piece symbol.

PR #203, opened 2017-02-06
enable FEN parser to work with multi-character piece symbols

alwey:
This patch extends the FEN parser in `Board::setFenString`. It supports multi-character piece symbols up to a length `maxsymlen` which depends on the predefined piece types. The longest match prevails, e.g.  the variant `Crazyhouse` uses `maxsymlen = 2` and so `Q~` (Pawn promoted to Queen) can be set instead of `Q` (Queen). This resolves issue #198.

Currently, apart from `Crazyhouse` and subvariants `maxsymlen`  equals one. Thus the reader will not be significantly slower.

Note: This extension only refers to pieces on the board, not to pieces in the reserve (in hand).
I hope this is useful. 

(updated patch and comment)

### DIFF
--- a/projects/lib/src/board/board.cpp
+++ b/projects/lib/src/board/board.cpp
@@ -57,6 +57,7 @@ Board::Board(Zobrist* zobrist)
 	  m_height(0),
 	  m_side(Side::White),
 	  m_startingSide(Side::White),
+	  m_maxPieceSymbolLength(1),
 	  m_key(0),
 	  m_zobrist(zobrist),
 	  m_sharedZobrist(zobrist)
@@ -114,7 +115,17 @@ void Board::initialize()
 		m_squares.append(Piece::WallPiece);
 	vInitialize();
 
+	m_maxPieceSymbolLength = 1;
+	for (const PieceData& pd: m_pieceData)
+		if (pd.symbol.length() > m_maxPieceSymbolLength)
+			m_maxPieceSymbolLength = pd.symbol.length();
+
 	m_zobrist->initialize((m_width + 2) * (m_height + 4), m_pieceData.size());
+}
+
+int Board::maxPieceSymbolLength() const
+{
+	return m_maxPieceSymbolLength;
 }
 
 void Board::setPieceType(int type,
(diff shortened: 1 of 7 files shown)


projects/lib/src/board/board.cpp
```diff
@@ -529,14 +534,23 @@ bool Board::setFenString(const QString& fen)
 		if (square >= boardSize)
 			return false;
 
-		pieceStr.append(c);
-		Piece piece = pieceFromSymbol(pieceStr);
-		if (!piece.isValid())
-			continue;
+		// read ahead for multi-character symbols
+		int l, lsym_minus1 = 0;
+		for (l = 0; l < maxsymlen && i + l < token->length(); l++)
```
ilaripih: It took me a while to understand this loop. To me it would be more logical to first try the maximum length token (because we have to try it eventually anyway), and if it doesn't match a piece symbol, then try `maxsymlen - 1` etc. Then we only have to call `setSquare()` when we really have the right piece, and we can break out of the loop early.

projects/lib/src/board/board.cpp
```diff
@@ -471,6 +471,11 @@ bool Board::setFenString(const QString& fen)
 		m_squares[i] = Piece::WallPiece;
 	m_key = 0;
 
+	int maxsymlen = 1;
```
alwey: Thank you. This is coming near a full circle. I already had this kind of solution in an earlier version, but then I did not calculate the necessary symbol length yet. I used a fixed length of one or two (per override in a protected method e.g. for Crazyhouse).  I will combine this according to your suggestion.

projects/lib/src/board/board.cpp
alwey: Your feedback yields the necessary fresh ideas. It is strange how I sometimes do not make or forget about obvious improvements.

alwey:
For `LoopBoard` (and subclass `ChessgiBoard`) there is an override method: maxPieceSymbolLength() which returns value 1. This prevents sneaking in the promoted piece types of Crazyhouse via FEN setup.

Perhaps cleaner: not to set those piece types in the constructor of `CrazyhouseBoard` if called during`LoopBoard` construction. This could be arranged by passing a boolean argument to the `CrazyhouseBoard` constructor. (?)

ilaripih:
Thanks for the changes. I actually like the current implementation where `LoopBoard` overrides the base class' `maxPieceSymbolLength()` method.

PR #219, opened 2017-06-02
support CECP V1 moves (xboard/gnuchessx old style moves)

alwey:
This is a proposal to support the alternative move format especially used by older xboard (CECP) engines.
Software collections often contain some or many older engines that do not work with cutechess.
Some variants have significant share of old engines.

While the prevalent move format for xboard engines is
`move e2e4` 
using algebraic notation  (or even using SAN `move e4`), the alternative format
`1. ... e2e4`
is still part of the specification, even for V2 engines.

This patch supports the alternative format 
NUMBER ... MOVE 	
in addition to the the prevalent format
move MOVE .

It has been tested with various engines and resolves problems with
_gnuchessx 6.1.1_
_phalanx XXII-pg_
which use the alternative format.

xboard communication format CECP is defined in:

http://www.open-aurec.com/wbforum/WinBoard/engine-intf.html

I hope this is useful.

### DIFF
--- a/projects/lib/src/xboardengine.cpp
+++ b/projects/lib/src/xboardengine.cpp
@@ -564,7 +564,8 @@ void XboardEngine::parseLine(const QString& line)
 
 		return;
 	}
-	else if (command.at(0).isDigit()) // principal variation
+	else if (command.at(0).isDigit()
+	     && !command.contains("."))	// principal variation
 	{
 		bool ok = false;
 		int val = 0;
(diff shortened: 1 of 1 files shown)


projects/lib/src/xboardengine.cpp
```diff
@@ -624,10 +627,14 @@ void XboardEngine::parseLine(const QString& line)
 			return;
 		}
 
-		Chess::Move move = board()->moveFromString(args);
+		// remove "..." of old format if necessary
+		int mark = (args.indexOf("..."));
+		const QString movestr = mark < 0 ? args : args.right(mark + 4);
```
ilaripih: You're lucky that the move string is exactly 4 characters long; otherwise this wouldn't work. `QString::right()` returns the `n` rightmost characters so `mark + 4` is a bit misleading because it's used as an index: http://doc.qt.io/qt-5/qstring.html#right

We already checked if `args.startsWith("...")` earlier so maybe could just use that info here to set `movestr = args.mid(4)`.

alwey: Ah, well spotted! This is an ugly mistake. It is hidden most of the time but promotions would call for trouble. As you wrote `mid(mark+4)` is what I intended to do. And also mark should always be zero in this branch of execution, so `mid(4)` should be the same.

PR #220, opened 2017-06-02
use edit command for xboard engines without setboard feature

alwey:
This is a proposal to support board setup for xboard engines that do not have the `setboard` feature. Instead of `setboard` the `edit` command is used to place the pieces.

The `edit` command is part of the CECP specification.

> The edit command is the old way to set up positions. For compatibility with old engines, it is still used by default, but new engines may prefer to use the feature command (see below) to cause xboard to use setboard instead.

http://www.open-aurec.com/wbforum/WinBoard/engine-intf.html

_Chessbase/Fritz_ extensions for en passant and castling `ep` and `castle` (see remarks at the end of CECP specification) are not supported yet.

I hope this is useful nevertheless.

### DIFF
--- a/projects/lib/src/board/board.cpp
+++ b/projects/lib/src/board/board.cpp
@@ -401,6 +401,24 @@ GenericMove Board::genericMove(const Move& move) const
 			   move.promotion());
 }
 
+QStringList Board::pieceList(Side side) const
+{
+	QStringList list;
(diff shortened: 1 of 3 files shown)


ilaripih:
Hey, I've wanted to do this for a while, so big thanks!

PR #223, opened 2017-06-07
support Janus Chess

alwey:
This is a proposal to support _Janus Chess_. This variant uses a 10x8 board and introduces the _Janus_, a piece with Knight and Bishop movement. This piece is  also known as _Princess, Archbishop, Hawk_ etc. in the context of other variants. As this piece is able to give mate on its own, the draw rate tends to be lower than in standard chess. It is similar to Capablanca Chess, but has different castling rules, different board setup, and one more Bishop+Knight piece per side instead of the _Chancellor_ (Rook+Knight).

Tests were done manually and with the engine Fairy-Max 5.0b .

I tried to abstract the notation symbol of the Janus "J" from the representation symbol on the graphical board scene. `JanusBoard` uses the "A" (Archbishop) graphical representation. In order to achieve this I made an extension to the `Board` class and changed one line in `BoardScene`.

Furthermore the castling rules of _Janus Chess_ were not compatible with the `WesternBoard` base class. I introduced a virtual method to override the castling files for the Kings in _Janus Chess_.

I hope this is acceptable nevertheless.

### DIFF
--- a/projects/lib/src/board/board.cpp
+++ b/projects/lib/src/board/board.cpp
@@ -131,12 +131,16 @@ int Board::maxPieceSymbolLength() const
 void Board::setPieceType(int type,
 			 const QString& name,
 			 const QString& symbol,
-			 unsigned movement)
+			 unsigned movement,
+			 const QString& gsymbol)
 {
 	if (type >= m_pieceData.size())
 		m_pieceData.resize(type + 1);
 
-	PieceData data = { name, symbol.toUpper(), movement };
+	const QString& graphicalSymbol = gsymbol.isEmpty() ? symbol : gsymbol;
+
+	PieceData data =
+		{ name, symbol.toUpper(), movement, graphicalSymbol.toUpper() };
 	m_pieceData[type] = data;
 }
 
@@ -181,6 +185,17 @@ QString Board::pieceString(int pieceType) const
 	if (pieceType <= 0 || pieceType >= m_pieceData.size())
 		return QString();
 	return m_pieceData[pieceType].name;
+}
+
+QString Board::representation(Piece piece) const
+{
+	int type = piece.type();
+	if (type <= 0 || type >= m_pieceData.size())
+		return QString();
+
+	if (piece.side() == upperCaseSide())
+		return m_pieceData[type].representation;
+	return m_pieceData[type].representation.toLower();
 }
 
 int Board::reserveType(int pieceType) const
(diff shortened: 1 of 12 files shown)


projects/lib/src/board/board.cpp
```diff
@@ -131,12 +131,15 @@ int Board::maxPieceSymbolLength() const
...
 			 const QString& symbol,
-			 unsigned movement)
+			 unsigned movement,
+			 const QString& gSymbol)
 {
 	if (type >= m_pieceData.size())
 		m_pieceData.resize(type + 1);
 
-	PieceData data = { name, symbol.toUpper(), movement };
+	const QString & gsym = gSymbol.isEmpty() ? symbol : gSymbol;
```
ilaripih: In our code convention the `&` symbol follows immediately after the type without a space between them.

projects/lib/src/board/board.h
```diff
@@ -309,11 +311,14 @@ class LIB_EXPORT Board
...
 		 *        piece can make.
+		 * \param gSymbol Select the piece's graphical representation.
+		 *	  If not set the \a symbol will be used (default).
 		 */
 		void setPieceType(int type,
 				  const QString& name,
 				  const QString& symbol,
-				  unsigned movement = 0);
+				  unsigned movement = 0,
+				  const QString & gSymbol = 0);
```
ilaripih: The default value of 0 is a bit weird when the type is a QString reference instead of a pointer. Am I correct to assume that the zero value invokes the `QString::QString(const char *str)` constructor? If that's the case, I'd rather use a default value of `QString()` to contruct a null string.

projects/lib/src/board/board.cpp
ilaripih: In our convention the `&` character comes immediately after the type without a space between them.

projects/lib/src/board/board.h
alwey: Yes, your suggestion is better and more clear.

projects/lib/src/board/board.cpp
alwey: Yes, this slipped through.

alwey:
New commit https://github.com/cutechess/cutechess/pull/223/commits/e59c93a3a4114f0104813f6734f2bfa4ebb1d8c2 includes the requested changes and is based on recent master.

artoj:
Is it possible to also write open the `gSym` variable? In Cute Chess single letter prefixes are usually used to meant for static / member variables and so on.

alwey:
Now `gSymbol` is replaced by `gsymbol`

artoj:
I meant that it should be written as `graphicalSymbol `or whatever. Abbreviations or other ways of shortening variable names usually just make the code harder to read.

ilaripih:
I agree with @artoj, `gSomething` usually implies a global variable, though in Cute Chess it would be `g_something` but still.

alwey:
Now `graphicalSymbol` is used again. This has been the first name I used while coding this.

PR #224, opened 2017-06-08
Support new mate score format of xboard engines in addition to old one

alwey:
The xboard (CECP) specification now maps (absolute) mate scores just above 100000 with mate distance in moves instead of plys.

> Mate scores should be indicated as 100000 + N for "mate in N moves", and -100000 - N for "mated in N moves".

I found that Fairy-Max 5.0b implements this new specification.
This patch adds support for the new mate scores. So it shows  +M9 instead of 1000.05 and -M10 instead of -1000.05. The scores out of (absolute) range 99900 to 100100 are unchanged. So far I have not seen any new problems with older mate score formats.

### DIFF
--- a/projects/lib/src/chessgame.cpp
+++ b/projects/lib/src/chessgame.cpp
@@ -40,6 +40,15 @@ QString evalString(const MoveEvaluation& eval)
 			str += "+";
 
 		// Detect mate-in-n scores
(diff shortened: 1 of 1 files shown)


ilaripih:
Thanks, looking good!

Issue #204, opened 2017-02-27
fails to compile on arm platform


flok99:
Trying to build cutechess-cli on the raspberry pi fails with:

gcc -c -pipe -std=c99 -O2 -fPIC -Wall -W -D_REENTRANT -DLIB_EXPORT= -DZ_PREFIX -DNDEBUG -DQT_NO_DEBUG -DQT_CORE_LIB -I/usr/lib/arm-linux-gnueabihf/qt5/mkspecs/linux-g++ -I. -Isrc -Icomponents/json/src -I3rdparty/fathom/src -isystem /usr/include/arm-linux-gnueabihf/qt5 -isystem /usr/include/arm-linux-gnueabihf/qt5/QtCore -I.moc -o .obj/tbprobe.o 3rdparty/fathom/src/tbprobe.c
3rdparty/fathom/src/tbprobe.c:33:23: fatal error: x86intrin.h: No such file or directory
 #include <x86intrin.h>
                       ^
compilation terminated.

Removing that include fixes the problem, at least for arm/rpi2.

ilaripih:
Thanks for the report. Obviously we should include `x86intrin.h` only on x86 platforms.

ilaripih:
Fixed.

PR #225, opened 2017-06-10
fathom tbprobe.c: include x86intrin.h only for x386-compatible builds

alwey:
This patch avoids problem #204.

### DIFF
--- a/projects/lib/3rdparty/fathom/src/tbprobe.c
+++ b/projects/lib/3rdparty/fathom/src/tbprobe.c
@@ -30,7 +30,9 @@
 #include "tbprobe.h"
 
 #ifdef __GNUC__
-#include <x86intrin.h>
+#  if defined(__i386) || defined(__x86_64__)
+#  include <x86intrin.h>
+#  endif
 #endif
 
 #define WHITE_KING              (TB_WPAWN + 5)


artoj:
Could this be applied to the upstream project and then we could pull the update?

alwey:
In the end this would be more effective of course. For now the patch will help non-x86-users of cutechess, so I should suggest to apply it.

I'll also ask @basil00 to include it upstream in order to avoid regressions.

alwey:
I just checked upstream there already is a related PR, 11 months old:
https://github.com/basil00/Fathom/pull/12

artoj:
Hmm.. can we just remove the header? We define `TB_NO_HW_POP_COUNT` in tbconfig.h

alwey:
Ah, good point. Are we missing anything?

ilaripih:
I've also added quite a few fathom patches directly to Cute Chess so I think we could remove the header completely; we're not going need a faster popcnt.

artoj:
Indirectly related to this: we should have some type of automatic ARM builds available as well. In general there isn't much machine dependent code in Cute Chess but still it would be good to be sure...

ilaripih:
I removed the header.

Issue #226, opened 2017-06-23
Crash when reading PGN


lucasart:
Crashes because variant "normal" is not recognized. I'm not sure if this PGN is standard compliant, and whether the tag `[Variant "normal"]` is correct. It was produced by Chess GUI, from Graham's tournament here:
http://kirill-kryukov.com/chess/discussion-board/viewtopic.php?f=7&t=9209
So, perhaps this is a bug for Chess-GUI. Anyway, I think it's good not to crash when we see an unrecognised tag, and simply assume the default (here standard chess).
```
[Event "The Demolisher"]
[Site "ChessGUI4"]
[Date "2017.06.22"]
[Round "1.1"]
[White "Rodent III 0.172 64-bit"]
[Black "Scorpio 2.7.7 64-bit"]
[Result "0-1"]
[Time "3:44:16 p.m."]
[ECO "E10"]
[Opening "Anti-Nimzo-Indian; E10"]
[TimeControl "40/1500:40/1500:40/1500"]
[PlyCount "134"]
[Number "1"]
[Termination "GUI adjudication"]
[BlackType "program"]
[WhiteType "program"]
[Variant "normal"]

1.d4 {[%eval 0,1] [%emt 00:00:00]} Nf6 {[%eval 0,1] [%emt 00:00:00]} 
2.c4 {[%eval 0,1] [%emt 00:00:00]} e6 {[%eval 0,1] [%emt 00:00:00]} 
3.Nf3 {[%eval 0,1] [%emt 00:00:00]} d5 {[%eval 0,1] [%emt 00:00:00]} 
4.g3 {[%eval 0,1] [%emt 00:00:00]} dxc4 {[%eval 0,1] [%emt 00:00:00]} 
5.Bg2 {[%eval 0,1] [%emt 00:00:00]} Bd7 {[%eval 0,1] [%emt 00:00:00]} 
6.Ne5 {[%eval 0,1] [%emt 00:00:00]} Bc6 {[%eval 0,1] [%emt 00:00:00]} 
7.Nxc6 {[%eval 0,1] [%emt 00:00:00]} Nxc6 {[%eval 0,1] [%emt 00:00:00]} 
8.O-O {[%eval 0,1] [%emt 00:00:00]} Qd7 {[%eval 0,1] [%emt 00:00:00]} 
9.e3 {[%eval 0,1] [%emt 00:00:00]} O-O-O {[%eval 0,1] [%emt 00:00:00]} 
[...]

PR #227, opened 2017-06-23
Map variant tag values chess and normal to standard variant for PGN i…

alwey:
This is a suggestion to improve PGN import compatiblity with some collections of game data by mapping the tags `[Variant "chess"] `and` [Variant "normal"]` to standard chess. This is related to issue #226

My observation is that neither the CLI nor the GUI crash, but issue a warning and ignore a game with `[Variant "normal"]`. Cutechess can import PGN files containing multiple (mixed) variants. 

Usually, for standard chess, the `Variant` tag should be omitted.
If a `Variant` tag is present, cutechess will only accept `standard` for standard chess.

I think it helps to map `normal` and `chess` to standard chess for better import compatibility. However, it does not help to accept any unknown `Variant` tags as standard chess. There are many variants that cutechess does not know (yet). We should not hide errors and compatibility problems but address them.

### DIFF
--- a/projects/lib/src/pgngame.cpp
+++ b/projects/lib/src/pgngame.cpp
@@ -157,6 +157,8 @@ bool PgnGame::parseMove(PgnStream& in)
 	if (m_moves.isEmpty())
 	{
 		QString tmp(m_tags.value("Variant").toLower());
+		if (tmp == "chess" || tmp == "normal")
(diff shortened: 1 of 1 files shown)


ilaripih:
The lack of standards is regrettable but we have to accept that, so this PR is fine. Thanks!

PR #241, opened 2017-07-21
support Andernach chess

alwey:
This is a suggestion to support **Andernach Chess**. Except for the kings a piece switches sides after it made a capture. A pawn capturing on the eighth rank first promotes and then switches sides. Apart from this standard chess rules apply. This variant originated from problem chess but can also be played in competition. A computer implementation has advantages against over the board play. The latter needs two sets of chess pieces and more interaction to exchange capturing pieces.

There seem to be no engines for this variant yet. Other than by playing myself in order to test the implementation I made some crude changes to the _sloppy_ engine.

While there I implemented boards for **Anti-Andernach Chess** (switch sides after non-captures) and  **Super-Andernach Chess** (always sitch sides after a move). These variants are more of interest for  problem chess than for competition.

I hope this is useful.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -39,6 +39,10 @@ is one of:
 Three-check Chess
 .It 5check
 Five-check Chess
+.It andernach
+Andernach Chess
+.It antiandernach
+Anti-Andernach Chess
 .It atomic
 Atomic Chess
 .It berolina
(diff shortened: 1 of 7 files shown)


ilaripih:
Thanks, I'll merge. I especially like the fact that you didn't need to modify the base classes.

PR #243, opened 2017-07-23
support Grid Chess

alwey:
This is a suggestion to support Grid Chess, a chess variant with an additional move restriction.

The board is devided into regions of 2x2 square. Moves within the regions are forbidden. So a piece cannot be captured or protected from inside its region. A king can not be checked from inside its region.

The implementation of `GridBoard` uses `RestrictedMoveBoard` as base class. This class can also be re-used for other variants.

_Grid Chess_ is accompanied by some related variants, where either the grid is displaced by file and rank or berolina pawns are used which more easily move between the grid's regions.

GridChessBoard has been tested manually and by engine-engine play with a hacked-up version of _Sloppy_.

The other variants have been tested manually.
I hope this is useful.

### DIFF
--- a/projects/lib/src/board/gridboard.cpp
+++ b/projects/lib/src/board/gridboard.cpp
@@ -1,0 +1,130 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


projects/lib/src/board/gridboard.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+	return "grid";
+}
+
+QString GridBoard::defaultFenString() const
+{
+	return "rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1";
+}
+
+bool GridBoard::leavesRegion(Move move, int fs, int rs) const
```
ilaripih: Looks like the `move` parameter could be a const reference instead of a value.

alwey: True, this saves a copy operation. Changed.

PR #242, opened 2017-07-21
support KnightMate

alwey:
This is a patch to support **Knightmate**. Here kings move like standard knights and vice versa. Castling moves are unchanged /wrt standard chess. 

I tested this implementation with some versions of _Fairy Max_ and _Sjaak II_. On the board the `King` is represented graphically by a `Knight` piece and the standard Knights are replaced by `Mann` or `Commoner` pieces that look like standard Kings.

PR #255, opened 2017-08-10
Support variant Two Kings Each (Wild9)

alwey:
This is a suggestion to support the chess variant _Two Kings Each_, also known as _wild9_, as per user request (discussion among other topics in #242). It is played on chess server ICC. 

Wild 9 is a variant with two Kings on each side. Only the King of either side that is currently closer to the a-file is royal and subject to check and mate. The other King is a normal piece and can be captured. Only the king starting on the e-file is able to castle. The castling rights are lost when any of the two kings of a side moves.

When two kings of a side are on the same file, the king closer to the first white rank is royal (i.e both sides refer to square a1), so the game has an asymmetry.

The `twokings` implementation `TwokingsEachBoard` has been tested with the existing engine pulsar2009-9b-64 and with the new engine by H. G. Muller, TK-max, a fairymax derivative. 

A symmetric subvariant has been proposed by @Nordlandia to reduce the white advantage, with black referring to the square a8 instead of a1. This subvariant is supported in `TwoKingsSymmetricalBoard` with ad-hoc variant name `sortland9`.

This variant has also been tested (albeit in a patched state to match the variant strings) with the help of TK-max and the white advantage seems to be less.

### DIFF
--- a/projects/lib/src/board/twokingseachboard.cpp
+++ b/projects/lib/src/board/twokingseachboard.cpp
@@ -1,0 +1,206 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "twokingseachboard.h"
+#include "westernzobrist.h"
+
+namespace Chess {
+
+TwoKingsEachBoard::TwoKingsEachBoard()
+	: WesternBoard(new WesternZobrist()),
+	  m_symmetrical(false),
+	  m_castlingSourceFile(4)
+{
+}
+
+Board* TwoKingsEachBoard::copy() const
+{
+	return new TwoKingsEachBoard(*this);
+}
+
+QString TwoKingsEachBoard::variant() const
+{
+	return "twokings";
+}
+
+QString TwoKingsEachBoard::defaultFenString() const
+{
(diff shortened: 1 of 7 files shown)


Nordlandia:
@alwey much appreciated for your work.

projects/lib/src/board/twokingseachboard.cpp
```diff
@@ -0,0 +1,206 @@
...
+*/
+
+#include "twokingseachboard.h"
+#include "westernzobrist.h"
+
+namespace Chess {
+
+TwoKingsEachBoard::TwoKingsEachBoard()
+	: WesternBoard(new WesternZobrist()),
+	  m_symmetrical(isSymmetrical()),
```
ilaripih: I think you forgot the `m_symmetrical` initialization here because it's also in `vInitialize()`. And of course virtual member functions shouldn't be called from the constructor anyway.

I think it's still a good idea to give `m_symmetrical` the default value in the constructor (to avoid undefined state/behavior) but without calling `isSymmetrical()`.

projects/lib/src/board/twokingseachboard.cpp
```diff
@@ -0,0 +1,206 @@
...
+		// find out which target square is needed by WesternBoard
+		Side side = sideToMove();
+		Move move1 = Move(kingSquare(side), target);
+		QString str2(lanMoveString(move1));
+		Move move2 = WesternBoard::moveFromLanString(str2);
+		Move move3 = Move(source, move2.targetSquare());
+		return move3;
+	}
+
+	return Move(move);
```
ilaripih: Would a simple `return move;` work?

projects/lib/src/board/twokingseachboard.cpp
alwey: OK, I think just like that. I wondered why the compiler ate this. When I coded this into the constructor I was fully expecting an error message or undefined behaviour at run time - but, no.

Thank you for the feedback. Being uneasy with the code myself, I will gladly change the value to a false in the constructor.

projects/lib/src/board/twokingseachboard.cpp
alwey: Yes, well spotted.

Nordlandia:
Illustration of wild9 checkmate with two kings. 

White is in check, but any king move will leave Kg2 in check, thereby checkmate. 

checkmate

alwey:
This no no mate position: K2f1 covers the royal King

Nordlandia:
More information on wild9 ->

Interesting modification - http://www.chessvariants.com/invention/kings

http://www.chessvariants.com/index/displaycomment.php?commentid=17040

PR #256, opened 2017-08-10
support Three Kings Chess

alwey:
This is a proposal to support _Three Kings Chess_. This is a variant quite different from _Two kings Each (Wild9)_.

In this variant, both sides have three kings and all of these are normal pieces. There is no check. The side that catches one of the opponent's kings wins. The starting position has a king in the middle of the back rank, and there are no rooks but kings at the corners. So the defense has to be balanced.

This variant has been tested by manual play.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -97,6 +97,8 @@ Symmetrized Wild 9
 Suicide Chess (Losing Chess Variant)
 .It superandernach
 Super-Andernach Chess
+.It threekings
+Three Kings Chess
 .It twokings
 Two Kings Each Chess (Wild 9)
 .It standard
--- a/projects/cli/res/doc/help.txt
+++ b/projects/cli/res/doc/help.txt
@@ -41,6 +41,7 @@ Options:
 			'sortland9': Symmetrical Two Kings Each Chess
 			'suicide': Suicide Chess (Losing Chess)
 			'superandernach': Super-Andernach Chess
+			'threekings': Three Kings Chess
 			'twokings': Two Kings Each Chess (Wild 9)
 			'standard': Standard Chess (default).
   -concurrency N	Set the maximum number of concurrent games to N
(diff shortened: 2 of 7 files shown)


Nordlandia:
Which of the three kings is royal?

By catche you mean checkmate?

Looks interesting.

alwey:
@Nordlandia ;
In this variant, both sides have three kings and all of these are normal pieces. There is no check.

Count your kings, if you have less than your opponent, you have lost ;-)

Nordlandia:
Please show starting position. 

Confusing at first.

alwey:
@Nordlandia :
>Please show starting position

"knbqkbnk/pppppppp/8/8/8/8/PPPPPPPP/KNBQKBNK w - - 0 1"

Nordlandia:
Just as described ;)

So neither king is royal, the game end once one of the king get captured or checkmated.

alwey:
@Nordlandia 
You might find this site interesting:  https://www.jsbeasley.co.uk/

This site also legally hosts the _Classified Encyclopedia of Chess Variants_ (2007) by late _B. D. Pritchard_, _ed. John Beasley_.
Navigate: (lhs) Chess Variants , then link (in center) "The Classified Encyclopedia of Chess Variants"
You can read or download the whole book, as well as a chapter wise representation.

Three Kings Chess is described in the Encyclopedia, chapter 10.2, page 82.


The book has 391 pages, including annotations in preparation for a 3rd edition  https://www.jsbeasley.co.uk/encyc/encyc.pdf.


>I shall not myself be producing an ECV 3 and anyone doing so may make use of what follows without further reference to me, but anyone wanting to attach David’s name to a new ECV must obtain permission from his family. It is of course open to anyone to produce a new encyclopedia independently, but I hope that anyone making significant use of David’s work (I do not see how a proper encyclopedia could now be written without making such use) will include a suitable acknowledgement.
(John Beasley)

PR #258, opened 2017-08-12
KnightMate using new symbols, fix for castling notation in KnightMate and Janus

alwey:
This PR is intended for the maintainance of some recent additions

 and comprises of 3 parts: 

1) I suggest to change the graphical representation of KnightMate:
King to use Royal Knight graphical symbol instead of Knight symbol.
Mann to use Mann symbol instead of King symbol.

km1

2) fix for Janus and Knightmate: I noticed that the SAN notation with trailing symbols, e.g.  O-O-O+ or O-O-O! was not processed correctly. 

3) remove some duplicate includes in boardfactory.cpp, They had slipped into master via my anti-0 branch.

HTH

### DIFF
--- a/projects/lib/src/board/boardfactory.cpp
+++ b/projects/lib/src/board/boardfactory.cpp
@@ -23,6 +23,7 @@
 #include "capablancaboard.h"
 #include "caparandomboard.h"
 #include "checklessboard.h"
+#include "chessgiboard.h"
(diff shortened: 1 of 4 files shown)


Nordlandia:
Looks good in my opinion. Also the shield seem to be proper aligned, not too big nor small.

PR #253, opened 2017-08-09
Separate UCI and Xboard high and mate scores

alwey:
Very high non-mate scores of UCI engines collide with the xboard mate score detection. Most engines are using lower scores, but e.g. _Houdini 5_ issues very high scores. For example, a UCI score of 99.99 gets translated to +M1, a false positive mate score. Reference: #250 

This patch moves most of the mate score preparation for xboard-scores from `ChessGame` into `XboardEngine`. The virtual mate scores now are mapped into the range around 98000 cp. `UciEngine` was adapted accordingly.

Remarks:
- While there I made a correction to the code for new Xboard scores.
- The choice 99000 for the virtual mate score was deliberate. It could be lower or higher. I also tried to use 200000 but then the output does not quite fit into the "Score" field on screen, but one could of course widen it.
- The vast majority of UCI and Xboard engines give scores which are correctly interpreted.
I have seen peculiar behaviour of some engines:

```
----------------------------
Matescores: 

1 sjeng 1 ply off: +M2 in M1 position.
1 nemorino  +M1 in +M3 position, 0.00 in win +M1: UCI win one move off, losing scores OK
H pulsar2009-9b-64 -M2 and -M1 in -M4 and -M2 positions: counts moves
2 baislicka -M12 -M8 -M4 in -M6, -M4, and -M2 positions
2 zetadva0301 -M4, -M8 in -M2, -M4 positions
2 gnucheese jumps +M18, +M14, +M10, +M6, +M2
2 olithink -M12, -M8, -M4
3 crafty -327.65 in -M2 position, 327.66 in +M1 position
3 crafty-25.2 -327.65 in -M2 position 327.66 in +M1 position
3 Gull3 -327.56 in -M2 position
3 Gull3 327.59 for +M1
8 fairymax 4.8Q -79.99 in -M2, -79.98 in -M4
8 shamax -79.99 in -M2 position
8 kingslayer -79.98 and -79.99 in -M4 and -M2 positions
* sjaakii -159.98 in -M2 position, 159.99 on UCI interface, problem created
* tunguska 299.99 for +M1 on UCI, problem created
0 amoeba-2.5.l64p omitted last losing score in -M2 position
0 hoichess-0.19.0 omits score in -M2 position
0 hoichess sometimes omits scores, sequence ok
5 heracles -100.07 in -M2 position
5 heracles UCI no mate score -100.09 is -M2
5 heracles UCI no mate score -100.15 is -M2
[...]

PR #261, opened 2017-08-17
The new output of the checkmate in the Cutechess GUI

sovaz1997:
In the position evaluation column, instead of displaying a very large score, the checkmate is displayed in a certain number of moves.

Screenshot:
screenshot_20170817_155511

### DIFF
--- a/projects/gui/src/evalwidget.cpp
+++ b/projects/gui/src/evalwidget.cpp
@@ -1,18 +1,18 @@
 /*
-	This file is part of Cute Chess.
+    This file is part of Cute Chess.
 
-	Cute Chess is free software: you can redistribute it and/or modify
-	it under the terms of the GNU General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
 
-	Cute Chess is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-	GNU General Public License for more details.
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
 
-	You should have received a copy of the GNU General Public License
-	along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 #include "evalwidget.h"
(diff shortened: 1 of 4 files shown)


alwey:
@sovaz1997 : It is a good idea to adapt the Evaluation Widget, at least fro the user's point of view.

However I would like to point out that the notation in the PGN file and on screen will be different.
The PGN file shows depth information in plies, e.g. +M7 -M6 -M5 -M4 +M3 -M2 +M1 in chess (and most variants). And in contrast, .. -M3, +M2, -M1 in losing chess variants.

The EvalWidget will show mate distances in moves, e.g.   #4, -#3, #3, -#2, #2, -#1, #1.
I think this will confuse some users and they will start to ask questions.

sovaz1997:
I agree that there is a defect in this. I think we need to know the opinion @ilaripih

ilaripih:
Thanks for the PR! Like @alwey said it would be better if the mate scores were displayed in the same format as in PGN, defined here: https://github.com/cutechess/cutechess/blob/master/projects/lib/src/chessgame.cpp#L42

I would also refactor the code so that we wouldn't have the same mate score logic in two places. Can you move that code from `chessgame.cpp`to the `MoveEvaluation` class, eg. `QString MoveEvaluation::scoreText() const`? And then use that method from `ChessGame` and `EvalWidget`.

sovaz1997:
It was necessary to leave the code evalString (from chessgame.cpp), as it displays not only an evalute score, but also other things (depth, time). I made the scoreText() method and use it in the evalString() and in the evalwidget.cpp.

screenshot_20170818_122319

sovaz1997:
I changed a little too much, I will now fix it

alwey:
This week I prepared a maintainance patch for may changes of the MoveEvaluation code in order to have less numbers that "magically" appear in the middle of source code (such like 99000 or 98800), as I am no big fan of such numbers.

Nordlandia:
+M29 do surely look better in my book. 

One question to OP @sovaz1997 -> how to make engine convert eval into mate in x?

sovaz1997:
Formule: score = MAX_SCORE - MOVES_TO_CHECKMATE
(Max detect: Checkmate in 1000:
I did the following in the UCI protocol:
```
if(score > MAX_SCORE + 1000 && score < MAX_SCORE - 1000) {
  std::cout << "cp " << score / PAWN_EV * 100;
} else if(score < 0) {
  std::cout << "score mate " <<  -MAX_SCORE - score) / 2;
} else {
  std::cout << "score mate " << (int)((WHITE_WIN - score + 1) / 2);
}
```

alwey:
@sovaz1997 : I would very much prefer plies as units - why? It is much easier to find bugs, because now you are throwing away information at the very last stage of the score processing by the division by 2. Please refer to PR #253, especially the listing of engines. I checked about seventy engines and found many with let's say - interesting deviations from the norm.

sovaz1997:
In my chess program it was like this, I wrote it here to make it clear, because there is no context. Corrected :)

alwey:
@sovaz1997 : 
> I did it the way I did in Stockfish, and the UCI did it that way.

Your're correct, the UCI mate distances are given in full moves. It is from the (traditional) Xboard protocol that we inherited the pseudo mate distances counted in plies. The new Xboard format is using moves for mate distance. I suggested not to change the step size also to stay in line with existing PGN files.

alwey:
When @sovaz1997 's fix is in place, we will be able to raise the pseudo mate score from 99000, to e.g. 1000000. I wanted a high score and only used 99000 because of the width of the score representation in the evaluation widget. This representation will be compact with the fix, and so it is no longer a problem. So another thumb up for @sovaz1997 !

ilaripih:
I made some final whitespace fixes and documented the `evalText()` method. The code is now merged.

Issue #210, opened 2017-04-07
Configurable default PGN output file for tournament games


fsmosca:
![image](https://cloud.githubusercontent.com/assets/22366935/26037032/61b84e12-391d-11e7-9578-9f4a25d80530.png)

alwey:
suggestion: show a Warning Dialog if no output file is given for a tournament and DefaultPGNFile is not set.

ilaripih:
We're going to add an option for default PGN output file to SettingsDialog's Tournament tab. I believe it would solve the problem, and it would not needlessly annoy those who don't want to save all tournament games to disk.

ilaripih:
Done by @alwey

rwbc:
Does this really work for tournament games?

I have specified a default pgn, but it seems no games of tournaments in case I forgot to specify a pgn file in the tournament
dialogue is saved there.

according line from my cutechess.ini
```
[games]
default_pgn_output_file=C:/test.pgn
```

Issue #208, opened 2017-03-20
Autosaving single engine games


Nordlandia:
Playing single engine games without saving it as .pgn (after game is finished) makes it impossible to restore game data, if you closes the gui. 

It should be an option to autosave to "Game Database" or desired directory. 

It's just a suggestion.

Cheers :)

alwey:
+1 for this enhancement

DefaultPGNFile _string_ (save to this file if set)

Somehow related to #210 (suggestion: show a Warning Dialog if no output file is given for a tournament and DefaultPGNFile is not set)

Nordlandia:
Yes, they are closely related. 

Approximately how long until next cutechess release at this point?

https://github.com/cutechess/cutechess/milestones

alwey:
problem is part (or duplicate) of #147

alwey:
@Nordlandia : The authors have shown their long-term dedication to this project. I assume that no updates can be expected when in conflict with real-life requirements.

ilaripih:
@alwey I think at this point we could consider you as one of the authors, you've contributed so much to this project. As for me and Arto, yes we're very busy but we'll still have our annual Cute Chess hackathon in July. I'd like to do a release before it, but if I don't, we'll definitely release a new version during the hackathon.

alwey:
@ilaripih: I feel honoured. Regarding a release, I also think that a minor (beta) release 0.9.x is interesting before the Hackathon. Actually, I have some code lying here for some months, e.g. `ShatranjBoard` and `CourierBoard `, among others. I should also work on a "final" _Losing Chess_ version (PR 200), the beta version of which has been tested by the Multi-Variant Stockfish/Fishtest developers fore some time.

Issue #147, opened 2016-09-17
Autosave for non-tournament games


ilaripih:
It should be possible to set a PGN file where all non-tournament games are saved automatically.

ilaripih:
Done by @alwey

PR #262, opened 2017-08-17
configurable default PGN files for single games and for tournament games

alwey:
This patch introduces settings into the configuration of cutechess GUI that specify a default PGN file name for non-tournament games (preset games.pgn) and a default tournament PGN file name (preset tournaments.pgn). They can be found in the General and the Tournament settings tabs.

The change helps to avoid loss of game information, especially of tournament games, if a user forgets to specify a tournament output file or to save a game directly.

References: issues #147 (#208), and  #210

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -725,6 +725,20 @@ void MainWindow::onGameFinished(ChessGame* game)
 			m_game = nullptr;
 		updateWindowTitle();
 	}
+
+	// save game notation of non-tournament games to default PGN file
+	if (!tab.m_tournament
+	&&  !game->pgn()->isNull()
+	&&  	(  !game->pgn()->moves().isEmpty()   // ignore empty games
+		|| !game->pgn()->result().isNone())) // without adjudication
+	{
+		QString fileName = QSettings().value("games/default_pgn_output_file", QString())
+					      .toString();
+
+		if (!fileName.isEmpty())
+			game->pgn()->write(fileName);
+			//TODO: reaction on error
+	}
 }
 
 void MainWindow::newTournament()
--- a/projects/gui/src/newtournamentdialog.cpp
+++ b/projects/gui/src/newtournamentdialog.cpp
@@ -271,4 +271,12 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
 void NewTournamentDialog::readSettings()
 {
 	ui->m_siteEdit->setText(QSettings().value("pgn/site").toString());
+
+	QString pgnName = ui->m_pgnoutEdit->text();
+	if (pgnName.isEmpty())
+	{
+		pgnName = QSettings().value("tournament/default_pgn_output_file",
+					    QString()).toString();
+		ui->m_pgnoutEdit->setText(pgnName);
+	}
 }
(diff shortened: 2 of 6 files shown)


projects/gui/src/newtournamentdialog.cpp
```diff
@@ -235,7 +235,11 @@ Tournament* NewTournamentDialog::createTournament(GameManager* gameManager) cons
 	t->setName(ui->m_nameEdit->text());
 	t->setSite(ui->m_siteEdit->text());
 	t->setVariant(ui->m_gameSettings->chessVariant());
-	t->setPgnOutput(ui->m_pgnoutEdit->text());
+	QString pgnName = ui->m_pgnoutEdit->text();
```
ilaripih: I think this is the wrong place to read the default PGN file path. Instead, `NewTournamentDialog::readSettings()` should have `ui->m_pgnoutEdit->setText(QSettings().value("pgn/default_tournament_pgn_file").toString());` so that the user could see and change the PGN output file before starting the tournament.

projects/gui/src/mainwindow.cpp
```diff
@@ -725,6 +725,20 @@ void MainWindow::onGameFinished(ChessGame* game)
...
 		updateWindowTitle();
 	}
+
+	// save game notation of non-tournament games to default PGN file
+	if (!tab.m_tournament
+	&&  !game->pgn()->isNull()
+	&&  	(  !game->pgn()->moves().isEmpty()   // ignore empty games
+		|| !game->pgn()->result().isNone())) // without adjudication
+	{
+		QString fileName = QSettings().value("pgn/default_pgn_file", QString())
```
ilaripih: I was actually planning on using the `pgn` namespace for PGN tags etc. I would rename this setting to `games/default_pgn_output_file` etc. I'm not sure what the namespace for non-tournament games should be, I guess `games` or `single_games` is okay. And there's already a `tournament` namespace so `pgn/default_tournament_pgn_file` would become `tournament/default_pgn_output_file` etc. I'd add the "output" part because games and tournament may also use PGN files as opening books (input).

projects/gui/src/settingsdlg.cpp
```diff
@@ -95,6 +111,42 @@ void SettingsDialog::browseTbPath()
 	dlg->open();
 }
 
+void SettingsDialog::browseDefaultPgnFile()
```
ilaripih: Can you copy the dialog settings from here: https://github.com/cutechess/cutechess/blob/master/projects/gui/src/newtournamentdialog.cpp#L85

Then the dialog title would be consistent ("Select PGN output file"), and `AcceptMode` should be `AcceptSave`; otherwise the user can't set a file that doesn't yet exist.

projects/gui/src/newtournamentdialog.cpp
alwey: ok, changed accordingly

projects/gui/src/mainwindow.cpp
alwey: used suggested structure

projects/gui/src/settingsdlg.cpp
alwey: Changed to suggested behaviour. In between the directory had been set to the filename in the line edit, so the directory argument incorrectly was given the filename. The Qt selector figured it out nevertheless.

PR #263, opened 2017-08-17
support Pocket Knight Chess

alwey:
This is a suggestion to support _Pocket Knight Chess_, a chess variant where each side has an extra knight in his pocket (reserve) which can be dropped onto an empty square instead of a normal move. So this is a very conservative type of drop chess which might appeal to people that are not familiar with drop chess. Pocket Knight Chess has been popular in The Netherlands and in Germany already at the beginning of the 20th century.

There are sub-variants of Pocket Knight with different set-up. There are remarks /wrt this in the code.
This implementation (main variant with the simplest rules) has been tested with _Sjaak II_, versions 1.3.1a and 1.4.1.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -87,6 +87,8 @@ Knightmate
 Loop Chess (Drop Chess Variant)
 .It losers
 Loser's Chess
+.It pocketknight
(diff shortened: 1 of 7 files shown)


ilaripih:
Looks good, thanks!

PR #265, opened 2017-08-18
support Co-regal Chess

alwey:
This patch supports _Co-regal Chess_, where the Queens are subject to check and mate just as the Kings. This implementation follows the original rules by Vernon R. Parton, i. e. there is no castling of Queens. It is quite compact and based on `WesternBoard`.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -57,6 +57,8 @@ Capablanca Random Chess
 Checkless Chess
 .It chessgi
 Chessgi / Drop Chess
+.It coregal
+Co-regal Chess
 .It crazyhouse
 Crazyhouse (Drop Chess Variant)
 .It displacedgrid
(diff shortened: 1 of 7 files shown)


Nordlandia:
I find Coregal Chess interesting. 

Royal Queens. 

white-rqueen
black-rqueen

Source: http://zoldsakk.hu/en/rules.php?v=4gcoreg

alwey:
Nice symbols. Thank you for the hint. I will incorporate these symbols RQ, rq  into my local wiki.svg to be safe.  If the PR got accepted, I think we could go with the normal Queen symbols for some time.

alwey:
I left the Queen Castling out, because this would complicate things. I think the Queens are mobile enough to be OK.

Nordlandia:
@alwey That's right. Neither queen does anything that the other one can't do

With many pieces remaining on the board, two queens may suffer from Larry Kaufman's principle of major piece redundancy.

Two queens in co-regal chess suffers from redundancy in addition to royality.

alwey:
@Nordlandia : concerns regarding threats increase very sharply with every new royal piece. More forks, skewers and pins to be avoided. Then it is just better to be happy with a new Rook instead of a Queen.

ilaripih:
Cute Chess is approaching the status of "variant king" among chess GUIs. Thanks!

alwey:
@ilaripih: Xboard and ChessV are having the lead.

Issue #251, opened 2017-08-08
store final position in EPD file


ilaripih:
For book and test suite makers it would be useful to store the final position of games in an EPD file.

ilaripih:
The cli part is completed, but I'll leave this open until we have the same setting in the GUI.

PR #266, opened 2017-08-19
EPD (or rather FEN) output of end position into a file (lib and cli)

alwey:
This patch helps to save the end positions of games into a specified file.  The notation is done in EPD (more exactly: FEN) format, one line per position.

This is part of #251

### DIFF
--- a/projects/lib/src/tournament.cpp
+++ b/projects/lib/src/tournament.cpp
@@ -81,6 +81,12 @@ Tournament::~Tournament()
 
 	delete m_openingSuite;
 	delete m_sprt;
+
+	if (m_pgnFile.isOpen())
+		m_pgnFile.close();
+
+	if (m_epdFile.isOpen())
+		m_epdFile.close();
 }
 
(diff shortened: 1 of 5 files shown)


projects/lib/src/tournament.cpp
```diff
@@ -250,6 +250,15 @@ void Tournament::setPgnCleanupEnabled(bool enabled)
 	m_pgnCleanup = enabled;
 }
 
+void Tournament::setEpdOutput(const QString& fileName)
+{
+	if (fileName != m_epdFile.fileName())
+	{
+		m_epdFile.close();
```
ilaripih: Probably the file handle should be closed in the destructor of `Tournament`. And the same goes for `m_pgnFile` which I've apparently forgotten to close.

alwey: Good point. I did not even check for leaks, just coded EPD alongside the PGN methods, so this has been out of my focus. Now open PGN and EPD-files are getting closed in the tournament destructor.

Issue #188, opened 2016-12-05
CLI engine match repeat opening N times


ddugovic:
Currently documentation suggests:
`-repeat		Play each opening twice so that both players get to play it on both sides`

I suggest enabling an opening to be repeated N times, which may be useful if using an opening book and selecting openings randomly:
`-repeat [N]		Play each opening twice (or N times) so that both players get to play it on both sides`

alwey:
This is about controlling the  sequence of game setups - also related: #95

ilaripih:
Now only the GUI part is left.

Issue #95, opened 2016-08-19
CLI option to not switch sides between games


ilaripih:
This is a feature request I got at talkchess.com: It should be possible to have engine X always play white or black.

ilaripih:
Done by @alwey

PR #269, opened 2017-08-21
new optional parameter n for -repeat, new option -noswap to not swap …

alwey:
…sides of paired engines.
 
This PR introduces a new optional parameter n to the `-repeat` option of cutechess-cli.

When set to a positive integer n, an opening will be played n times during an encounter between two engines. If the parameter is omitted, as before the opening will be played twice (default). The program issues a warning if the number of games per encounter and the number of opening repetitions do not match well, e.g. `-games 4 -repeat 3` playing with 3 repetitions when 4 games are played per encounter.
The repetitions cannot extend over several rounds. A new round will start with a new opening.
Requested in #188.

The PR also introduces a new option `-noswap`.

The engines usually swap sides when having an encounter of more than one game. When the option `-noswap` is used, the engines will further play the sides on which they started the round and not switch.
Feature suggested in #95.

The PR only adds functionality to the CLI, whereas the formal change to the GUI is non-functional.

I hope this helps.

### DIFF
--- a/projects/lib/src/tournament.h
+++ b/projects/lib/src/tournament.h
@@ -196,12 +196,19 @@ class LIB_EXPORT Tournament : public QObject
 		void setEpdOutput(const QString& fileName);
 
 		/*!
-		 * Sets the opening repetition mode to \a repeat.
+		 * Sets the number of opening repetitions to \a count.
 		 *
-		 * If \a repeat is true, each opening is repeated for two
-		 * rounds; otherwise each game gets its own opening.
+		 * Each opening is played \a count times before
+		 * going to the next opening. The default is 1.
 		 */
-		void setOpeningRepetition(bool repeat);
+		void setOpeningRepetitions(int count);
+		/*!
+		 * Sets the side swap flag to \a enabled.
+		 *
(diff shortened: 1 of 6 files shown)


projects/lib/src/tournament.h
```diff
@@ -196,12 +196,19 @@ class LIB_EXPORT Tournament : public QObject
...
-		 * Sets the opening repetition mode to \a repeat.
+		 * Sets the number of opening repetitions to \a count.
 		 *
-		 * If \a repeat is true, each opening is repeated for two
-		 * rounds; otherwise each game gets its own opening.
+		 * Each opening is played \a count times before
+		 * going to the next opening. The default is 1.
 		 */
-		void setOpeningRepetition(bool repeat);
+		void setOpeningRepetitions(int count = 1);
```
ilaripih: This function is always called with an explicit count and I don't see why it would be called without it, so the default parameter value of 1 is not needed. The documentation ("The default is 1") should still remain.

ilaripih:
I closed the related issues because there were only requested for cli. If they're needed in GUI then we can create new issues.

Issue #211, opened 2017-05-06
Feature: Allow manual adjudication


fsmosca:
Termination tag: adjudication
Termination marker {User adjudication}

ilaripih:
Cute Chess already has the Tags table which I'm planning on making editable. But yes, adjudication should have its own button or menu item.

ilaripih:
Was done a while ago, forgot to close issue.

Issue #239, opened 2017-07-21
Predicted move & accuracy in percentage


Nordlandia:
Thanks a lot for releasing v1.0.0 🌻

Pondermove & ponderhit work only if permanent brain is activated. Ponder comes at 50% CPU penalty for each engine, on single CPU ship. Thus many engine testers avoid using ponder. 

Predicted move accuracy functions same way but it doesn't require ponder to be enabled. 

This is meant as an improvement. 

This feature is implemented in Fritz GUI, - in ongoing engine matches, one can see the percentage. 

![123](https://user-images.githubusercontent.com/26534454/28472501-16a58896-6e41-11e7-8fb0-8dbc9c0f9ae4.png)

Nordlandia:
Alternatively ponder move & percentage should work if ponder is not enabled.

ilaripih:
Yes, I can see that this would be useful. Most UCI engines print the ponder move even when pondering is disabled. And if they don't or if the engine uses Xboard protocol, we can get the predicted move from the last PV line.

ilaripih:
Pondermove and ponderhit are now shown for UCI engines even when pondering is disabled.

Nordlandia:
I tried update MSYS2 installer and new compiled GUI don't display 😢

ilaripih:
The changes were in cutechess-lib. It's possible that you have to completely recompile it before linking to cutechess-gui.

Nordlandia:
I'm following _making a release_ recipe, which turns out don't work (doesn't include latest implemention)

Recompile msys64 content is new to me.

https://github.com/cutechess/cutechess/wiki/Making-a-release

ilaripih:
Let's reopen this, it looks like my implementation of predicted move is not correct. I tested with Stockfish which always sends the ponder move, but other engines may not do that if pondering is disabled.

Nordlandia:
CuteChess report pondermove but unfortunately it's the previous move. It looks like the tab don't update once one side has made it's move. It report previous move and sometimes get stuck on that for a while.

PR #270, opened 2017-08-24
Game adjuducation by user

alwey:
This patch adds three menu items to the "Game" menu. The user can adjudicate an _ongoing_ game as Draw, White Win, and Black Win: a call to a new slot on the `ChessGame` class stops and adjudicates the game displayed in the current game tab (single game or tournament game).

Refer to #211 (for ongoing games)

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -159,6 +159,10 @@ void MainWindow::createActions()
 
 	m_flipBoardAct = new QAction(tr("&Flip Board"), this);
 
+	m_adjudicateDrawAct = new QAction(tr("Ad&judicate Draw"), this);
+	m_adjudicateWhiteWinAct = new QAction(tr("Adjudicate Win for White"), this);
+	m_adjudicateBlackWinAct = new QAction(tr("Adjudicate Win for Black"), this);
+
 	m_quitGameAct = new QAction(tr("&Quit"), this);
 	m_quitGameAct->setMenuRole(QAction::QuitRole);
 	#ifdef Q_OS_WIN32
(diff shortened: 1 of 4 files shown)


Nordlandia:
Wonderful addition.  

Please consider #239 as well.

projects/gui/src/mainwindow.cpp
```diff
@@ -159,6 +159,10 @@ void MainWindow::createActions()
 
 	m_flipBoardAct = new QAction(tr("&Flip Board"), this);
 
+	m_adjudicateGameAct2 = new QAction(tr("Ad&judicate Draw"), this);
```
ilaripih: Could you rename these to `m_adjudicateDrawAct`, `m_adjudicateWhiteWinAct` and `m_adjudicateBlackWinAct`?

projects/gui/src/mainwindow.cpp
```diff
@@ -1034,6 +1047,43 @@ bool MainWindow::askToSave()
...
+{
+	adjudicateGame(1);
+}
+
+void MainWindow::adjudicateGame2()
+{
+	adjudicateGame(2);
+}
+
+void MainWindow::adjudicateGame(int value)
```
ilaripih: Can you create an enum for the `value` parameter, eg. `MainWindow::UserAdjudication` etc?

projects/gui/src/mainwindow.cpp
alwey: yes, of cause this is clearer

projects/gui/src/mainwindow.cpp
alwey: yes, cleaner code

Issue #213, opened 2017-05-14
Don't allow doubles in tour's players list


fsmosca:
![image](https://cloud.githubusercontent.com/assets/22366935/26036995/65ca04f6-391c-11e7-90d2-b1d76abdd33f.png)

alwey:
No, actually this is no bug but a feature:

Actually the list is a list of engine configurations rather than engines. You can set up a tournament , say with 3 `Rodent III 0.172 x64 1 CPU` engines. Then modify the entries (button w/ wrench symbol) and set up special parameters (this is using setoption for UCI engines). You can rename these engine settings like EXP1, EXP2  or AGGRO1.

fsmosca:
I know how to setup a proper tournament. I just want cutechess to be smart, not allowing same engine name in the tournament list. This is especially useful when running more than 20 engines where visual examination would already take some effort. Perhaps a box telling the user that there are same engine names in the tournament list after the user presses the OK button would be fine.

alwey:
Yes, a hint of some kind, a button or a dialog box. Indeed, this would be helpful.

ilaripih:
This was done a while ago, forgot to close the issue.

PR #271, opened 2017-08-24
Do not start tournament with duplicate configuration names

alwey:
This patch prevents the user from starting a tournament with non-unique configuration names. If there are ambiguities in the engine list the `NewTournamentDialog::canStart()` method returns false and the start button gets disabled. Furthermore the button shows a "Resolve Duplicates!" text as long as there are duplicates in the list.
Resolution can be accomplished by removing or renaming entries of engine configurations.

Reference: Request #213

### DIFF
--- a/projects/gui/src/newtournamentdialog.cpp
+++ b/projects/gui/src/newtournamentdialog.cpp
@@ -169,6 +169,9 @@ void NewTournamentDialog::configureEngine(const QModelIndex& index)
 
 	if (dlg.exec() == QDialog::Accepted)
 		m_addedEnginesManager->updateEngineAt(row, dlg.engineConfiguration());
+
+	QPushButton* button = ui->buttonBox->button(QDialogButtonBox::Ok);
+	button->setEnabled(canStart());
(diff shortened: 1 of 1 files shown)


projects/gui/src/newtournamentdialog.cpp
```diff
@@ -192,6 +195,17 @@ bool NewTournamentDialog::canStart() const
 	if (m_addedEnginesManager->engineCount() < 2)
 		return false;
 
+	QPushButton* button = ui->buttonBox->button(QDialogButtonBox::Ok);
+
+	// check for duplicate configuration names
+	if (m_addedEnginesManager->engineNames().count()
+	!=  m_addedEnginesManager->engineCount())
+	{
+		button->setText("Resolve Duplicates!");
```
ilaripih: Can you wrap the "Resolve Duplicates!" text inside `tr()` so that it can be localized?

alwey: ah, good catch!

PR #277, opened 2017-09-01
mark duplicates in engine configuration list

alwey:
This patch shows duplicate entries in the list of tournament engine configurations. Non-unique entries are marked grey (the user has to resolve the duplicates by renaming or removing entries).

This is not optimised for speed. However, lists with several hundred entries are no problem on my modest laptop.

This PR refers to issue #213 and PR #271.


ilaripih:
Thanks for the PR! Yes, performance-wise this is not optimal but I don't think users can notice that.

PR #282, opened 2017-09-23
output of end positions of tournament games into a file (gui)

alwey:
This patch helps to save end positions of tournament games in FEN format, one position per line when using the GUI.

This is part of #251 and complements PR #266.


ilaripih:
Looking good, thanks!

PR #290, opened 2017-10-29
Eight piece sets by Maurizio Monge

alwey:
This PR includes eight SVG piece sets for standard chess by Maurizio Monge published [here](http://poisson.phc.unipi.it/~monge/chess_art.php). The piece sets are already in use by Wikipedia, Xboard, PyChess and others. I adapted the SVG files to use CuteChess compatible index strings and added license metadata for [CC-BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/), which is GPL V3 one-way-compatible.

_Remark to users_: To use the piece set with cutechess GUI one has to wait for a GUI patch for piece selection. 

Adventurous users may try to rename `default.svg `to `default.svg.orig` and copy a new piece set to `default.svg` instead (within directory `projects/gui/res/chessboard`). Then recompile the GUI. The sets only contain the 6 standard piece types, so you will see void squares when trying out Crazyhouse or Capablanca chess, for example.

HTH

sp1

### DIFF
--- a/projects/gui/res/chessboard/MaurizioMonge/celtic.svg
+++ b/projects/gui/res/chessboard/MaurizioMonge/celtic.svg
@@ -1,0 +1,1633 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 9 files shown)


ilaripih:
Thanks a bunch! I think it would be cool to create Capablanca pieces for these piece sets (I did that for the default set with Inkscape and it turned out okay). But I'll definitely merge these for later use.

PR #295, opened 2017-11-04
Use better name on interface for Symmetrical Two Kings Each variant

alwey:
#The id `sortland9` for the symmetrical version of _Two Kings Each Chess_ (ICC Wild/9) has been chosen ad-hoc. Recently _Two Kings Each Chess_ (`twokings`) has been added to Multi-Variant [Stockfish](https://github.com/ddugovic/stockfish/) and the symmetrical version will also be included. It was [suggested](https://github.com/ddugovic/stockfish/issues/431) to use `twokingssymmetric`, a more chess-related variant identifier. As far as I know, no third party engines are affected by this change.

This id change would help testing the symmetrical variant in multi-variant Fishtest, too.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -97,8 +97,6 @@ Pocket Knight Chess
 Racing Kings Chess
 .It slippedgrid
 Slipped Grid Chess
-.it sortland9
-Symmetrized Wild 9
 .It suicide
 Suicide Chess (Losing Chess Variant)
 .It superandernach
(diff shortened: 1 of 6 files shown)


Nordlandia:
I'm very glad they added TwoKings into testing framework. I proposed _TwoKings_ for lichess some time ago  see link at bottom, but it got downvoted quite badly. 

_twokingssymmetric_ is better name and self explanatory at same time. 

https://github.com/ornicar/lila/issues/3442

PR #319, opened 2017-11-23
fix inaccuracy in JanusBoard::moveFromSanString

### LINKED ISSUE #316: Failing test in debug mode
When running in debug mode (tested under Linux, gcc and clang), I see a failing Q_ASSERT.

```
QFATAL : tst_Board::moveStrings(janus castling san1) ASSERT: "!move.isNull()" in file src/board/board.cpp, line 665
FAIL!  : tst_Board::moveStrings(janus castling san1) Received a fatal error.
```
The line in question is:
```
Q_ASSERT(!move.isNull());
```
I was unable to figure out why move is null.

alwey:
resolves #316

### DIFF
--- a/projects/lib/src/board/janusboard.cpp
+++ b/projects/lib/src/board/janusboard.cpp
@@ -92,13 +92,13 @@ Move JanusBoard::moveFromSanString(const QString& str)
 	if (hasCastlingRight(sideToMove(), KingSide))
 	{
 		Move castlingMove = WesternBoard::moveFromSanString("O-O");
-		if (str == sanMoveString(castlingMove))
+		if (!castlingMove.isNull() && str == sanMoveString(castlingMove))
 			return castlingMove;
 	}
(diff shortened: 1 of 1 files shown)


projects/lib/src/board/janusboard.cpp
```diff
@@ -92,13 +92,13 @@ Move JanusBoard::moveFromSanString(const QString& str)
 	if (hasCastlingRight(sideToMove(), KingSide))
 	{
 		Move castlingMove = WesternBoard::moveFromSanString("O-O");
-		if (str == sanMoveString(castlingMove))
+		if (castlingMove != Move() && str == sanMoveString(castlingMove))
```
ilaripih: I think that instead of `castlingMove != Move()` you could write `!castlingMove.isNull()`

alwey: Thank, you. This is much better. Changed accordingly.

PR #324, opened 2017-12-09
Fix two kings insufficient material adjudication

ddugovic:
Include commoner(s) in the material count (valued as a major piece).

@ianfab This is the `cutechess-cli` I have used in my recent testing.


ilaripih:
Thanks for the contribution!

Issue #328, opened 2017-12-24
Minimal PGN export missing Variant tag for variant games


ddugovic:

alwey:
@ddugovic:  Confirmed. In minimal PGN format only the main seven tags are written. This is not feasible for variant games. A fix will be available soon.

ilaripih:
Thanks for the report @ddugovic and thanks for the fix @alwey

PR #329, opened 2017-12-24
This patch adds the Variant tag to the minimal PGN output.

alwey:
Fixes issue #328.


projects/lib/src/pgngame.cpp
```diff
@@ -279,6 +279,12 @@ bool PgnGame::write(QTextStream& out, PgnMode mode) const
 		writeTag(out, "SetUp", m_tags["SetUp"]);
 	}
 
+	if (mode == Minimal && m_tags.contains("Variant")
+	&&  variant() != "standard")
```
alwey: The method isStandard() does not work here in case of FEN setup.

Issue #121, opened 2016-09-05
Accept case-insensitive variant name in PGN


ilaripih:
When parsing a PGN file there are no standard names for a lot of variants so we shouldn't be too strict. For example Lichess uses `Crazyhouse` instead of `crazyhouse`.

ilaripih:
Done

PR #330, opened 2017-12-25
Resolve problems when importing PGN data containing capital Variant tag values

alwey:
like "Antichess" into the game database.

The GUI crashes with segmentation fault when importing a PGN game with some Capital letters in its Variant tag. This only occurs when the "Variant" tag matches a supported variant in lower case.

This patch supports the work done on #121.


ilaripih:
Thanks a bunch!

PR #331, opened 2017-12-26
support Seirawan Chess (a.k.a. S-Chess)

alwey:
This PR is a suggestion to support _Seirawan Chess_ or _S-Chess_.

In 2007 Yasser Seirawan and Bruce Harper introduced a new variant of Standard Chess. Their intention was to reduce the draw rate by having Capablanca pieces but using the common 8x8 board. At start the standard chess set up is used. However, each side has in reserve a Knight+Bishop combination, called _Hawk_ (Archbishop in Capablanca Chess) and a Knight+Rook combination, called _Elephant_ (Capablanca Chess: Chancellor). One of these pieces has the option to enter the board on a square of the own base rank immediately after this square has been vacated for the first time. No piece can enter on a square that has been vacant before the move. Only one piece can enter per move. There is no obligation to use the reserve pieces at all.

se1ase2a
_Selection of reserve piece to be inserted on square b8 which has been vacated by the knight moving to c6._

`SeirawanBoard` is derived from `WesternBoard` and uses support already in place for variants with drop moves. However, _Seirawan Chess_ has a different character and needs some adjustments.

The channeling moves have been coded on the engine interface and also internally as moves with promotion to reserve pieces. This format has already has been used by Seirawan Chess engines and _Xboard_ when communicating via _CECP_, and recently also by Stockfish via UCI . As a consequence for the Cutechess GUI  a player can pick the desired piece to be inserted from the choser when making a move from the base rank.

The GUI has been extended in order to support optional promotions, so one can not only chose a promotion type but click somewhere outside the piece choser in order to select no piece at all.

The GUI also has been adapted to mirror transition sequences of making moves and undoing moves. This difference is relevant for composite moves (like channeling moves or castling with channeling).

The implementation has been tested with various Seirawan Chess engines: Sigla v2.1, xleo-83, Sjaak II 1.3.1 and 1.4.1, Fairymax 5,0b, and "Stockfish for Seirawan Chess".

Also a number of test cases have been added.
HTH

se3a
se4a
_Selection of reserve piece to be inserted at either e1 or h1 when castling short. There are four choices: two different pieces times two target squares. Here the Elephant was selected for insertion at the rook square h1._

### DIFF
--- a/projects/lib/src/board/seirawanboard.cpp
+++ b/projects/lib/src/board/seirawanboard.cpp
@@ -1,0 +1,438 @@
+/*
+    This file is part of Cute Chess.
+
+    Cute Chess is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Cute Chess is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Cute Chess.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "seirawanboard.h"
+#include "westernzobrist.h"
+#include "boardtransition.h"
+
+namespace Chess {
+
+SeirawanBoard::SeirawanBoard()
+	: WesternBoard(new WesternZobrist())
+{
+	setPieceType(Hawk, tr("hawk"), "H", KnightMovement | BishopMovement, "A");
+	setPieceType(Elephant, tr("elephant"), "E", KnightMovement | RookMovement, "C");
+
+	// following "virtual" piece types are only used for castling/channeling disambiguation
+	setPieceType(rookSquareChanneling(Hawk, forward), tr("auxhawk"), "X", 0, "A");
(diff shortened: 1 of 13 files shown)


projects/lib/src/board/seirawanboard.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+void SeirawanBoard::generateMovesForPiece(QVarLengthArray< Move >& moves,
+					  int pieceType,
+					  int square) const
+{
+	if (!isValidSquare(chessSquare(square)))
+		return;
+
+	QVarLengthArray< Move > moves1, moves2;
+	moves1.clear();
```
ilaripih: Aren't `moves1` and `moves2` already empty or cleared after construction?

projects/lib/src/board/seirawanboard.cpp
```diff
@@ -0,0 +1,440 @@
...
+
+		if (side == upperCaseSide())
+			s.append(squareString(i).mid(0,1).toUpper());
+		else
+			s.append(squareString(i).mid(0,1).toLower());
+	}
+
+	QStringList vfen = vfs.split(' ', QString::SkipEmptyParts);
+
+	if ((vfen.count() > 0 && vfen.at(0)!= "-") || s.length() == 0)
```
ilaripih: In Qt the convention is to use `QStringList::isEmpty()` or `QString::isEmpty()` to test for emptiness.

alwey: ah, thank you, will be changed

projects/lib/src/board/seirawanboard.cpp
alwey: Well spotted.

Issue #332, opened 2018-01-05
timestamp


flok99:
Hi,

Can you maybe add at full timestamp (e.g. including the hour/min/sec) in the pgn output or give an other means of mapping game results to logging of the engine?

Situation: a chess engine causes invalid moves. As 1000ths of games are played during the night, it is impractical to map game to a location in the logfile. As the engine logfile contains timestamps, then that could be a key.

alwey:
+1 for optional comment in PGN at game end.

ilaripih:
Maybe it would be easiest to parse if the timestamp was a new PGN tag. Is the "game started" timestamp enough or would you need a "game ended" timestamp as well?

alwey:
I have got something like that. A problem may be to switch off the timestamp tags.

```
[Event "?"]
[Site "?"]
[Date "2018.01.13"]
[Round "?"]
[White "crafty"]
[Black "Dirty"]
[Result "0-1"]
[ECO "A00"]
[GameEndTime "2018-01-13T19:34:09.204 CET"]
[GameStartTime "2018-01-13T19:32:59.310 CET"]
[Opening "Dunst (Sleipner, Heinrichsen) Opening"]
[PlyCount "116"]
[TimeControl "30+0.3"]

1. Nc3 {+0.38/15 0.74s} c5 2. Nf3 {+0.72/14 0.73s} Nc6 3. e4 {+0.47/14 1.1s} d6 
...

```

ilaripih:
Thanks for the good work @alwey

PR #336, opened 2018-01-14
PGN timestamp tags

alwey:
This PR supports #332. I am not sure everybody likes the tags, so it might be better to make the tags optional.


ilaripih:
Awesome, thanks! If people don't want non-standard PGN tags they can use minimal PGN mode which already exists.

PR #337, opened 2018-01-14
Fix deletion of pgn data before signal emission

gvreuls:
Fixes a bug in the library where tournament pgn data are deleted before
the gameFinished signal is emitted, denying handlers access to data.
The fix deletes the pgn data immediately after sending the signal
instead.


ilaripih:
Thanks for fix, looks good.

PR #338, opened 2018-01-14
Fix automatic creation of ECO tags while reading EcoNode file

gvreuls:
Fixes a bug in the library where PgnGame::read() automatically creates
the opening tags (ECO, Opening, and Variation) while reading a file that
already contains those tags.  This creates a bug in
EcoNode::initialize(PgnStream&) which uses PgnGame::read() to create the
EcoNode tree from a PGN file.


ilaripih:
Good stuff, thanks!

Issue #137, opened 2016-09-11
Request to choose Merida chess pieces and different color squares


kervinck:
The current pieces are the Wikipedia/CBurnett pieces, I think. They are beautiful, but personally I find the thick rooks a bit annoying. I prefer the classic Merida. It would be nice to have a way to select from the common classic fonts.

I think there a 6 commonly used sets that are also timeless and free for use, in alphabetical order:
- Alpha
- CBurnett/Wikipedia
- Cheq
- Leipzig
- Merida
- USCF
Preview (.png) and references to original sources on https://marcelk.net/chess/pieces/ (UCSF pieces still missing there, but I believe they are equally good). 

It would be really nice to be able to bundle at least these 6 classics in non-transparent, plain rendering (no distracting shading/shadows and such).

Same for the colours. Some prefer blue hues, etc.. Suggestion to provide a couple of well-chosen presets.

Obviously this is not the most urgent feature request.

ilaripih:
It's actually not so straightforward to use a new piece set because Cute Chess needs them in SVG (scalable vector graphics) format in a single image where pieces are identified by an SVG id. And because in addition to standard pieces we also need the special pieces for Capablanca and Crazyhouse variants.

kervinck:
I will also see if I can assist in providing compatible SVG versions.

Although they look good on all resolutions and the file sizes are small, my PNG renderings were intended to be used in older versions of XBoard. Not as a suggestion to use where SVG is also available.

alwey:
I compiled an indexed set of SVG  Merida pieces from open source (wikipedia and PyChess) and created missing pieces `Q~,R~,B~,N~,q~,r~,b~,n~,A,C,a,c` from original basic pieces with inkscape. To be safe I have to investigate the origin and the process that led to the basic images for credit and license reasons.

I tested the SVG piece set with my local build.

alwey:
Update:

Technically I have prepared the `Merida` SVG chess set with 12x2 pieces (standard plus Carrera pieces plus Crazyhouse pieces). One issue exists yet regarding the license chain of part of the pieces.

In addition, I prepared all of the fine _Maurizio Monge_ standard piece sets (6x2 pieces). For legal reasons the SVGs of the artwork have to be enhanced with author and license tags yet.

The CBurnett Wiki default piece set currently comprises of 20x2 pieces for variant play.

alwey:
Uploaded Maurizio Monge sets: #290.

PR #342, opened 2018-01-19
Merida pieces

alwey:
This PR contains a SVG file with piece graphics derived from the original True Type Font by Armando H. Marroquin.

Derivative art work by Felix Kling has been used with his permission by the PyChess project.
I arranged the SVG single piece files into one file and added new piece graphics of Carrera (Capablanca) pieces and promoted pieces used e.g. for Crazyhouse.

This is intended to be released under license CC-BY-SA-4.0 international, which is designed to be compatible for usage with GNU General Public License V3.

I would like to get feedback from Felix Kling and Tamás,  @gbtami, to be sure they do not object. 
EDIT: This relates to user request #137 .

merida1

### DIFF
--- a/projects/gui/res/chessboard/chessboard.qrc
+++ b/projects/gui/res/chessboard/chessboard.qrc
@@ -1,6 +1,7 @@
 <!DOCTYPE RCC><RCC version="1.0">
 <qresource>
 	<file>default.svg</file>
+	<file>merida.svg</file>
 	<file>MaurizioMonge/celtic.svg</file>
 	<file>MaurizioMonge/eyes.svg</file>
 	<file>MaurizioMonge/fantasy.svg</file>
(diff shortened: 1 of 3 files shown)


gbtami:
Huh. I'v just realized that I never asked Felix Kling permission to use Merida in PyChess :( I will try to contact him and correct my mistake.

alwey:
@gbtami : I would be glad if you could also relay our case to Felix. If everything is ok, then feel free to use the new pieces as well.

gbtami:
@alwey sure. Already relayed to @DeepKling

DeepKling:
So to you as well: Feel free to use these pieces :-) .

Just to add: There is a new version of that piece set: https://github.com/ornicar/lila/files/1177425/Merida15.zip . See https://github.com/ornicar/lila/issues/3318 for pros and cons of using the "improved" version of this piece set :-) .

alwey:
@gbtami : @DeepKling : Thank you very much - vielen Dank!

ilaripih:
Great stuff, thanks!

PR #343, opened 2018-01-19
Add function to resign a game via GUI menu item

alwey:
This adds a menu item to the "Game" menu of the GUI with the functionality to resign a game for a human player. When the new item "Resign" is selected, the side to move resigns the game. However, if the side to move is not a human (local) player then the side to wait resigns the game. If both sides are played by engines, no resignation will be triggered - and the menu item will be rendered inactive anyway.

In addition, the menu items for game adjudication are disabled if the game shown in the active tab is finished.

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -163,6 +163,8 @@ void MainWindow::createActions()
 	m_adjudicateWhiteWinAct = new QAction(tr("Adjudicate Win for White"), this);
 	m_adjudicateBlackWinAct = new QAction(tr("Adjudicate Win for Black"), this);
 
+	m_resignGameAct = new QAction(tr("Resign"), this);
+
 	m_quitGameAct = new QAction(tr("&Quit"), this);
 	m_quitGameAct->setMenuRole(QAction::QuitRole);
 	#ifdef Q_OS_WIN32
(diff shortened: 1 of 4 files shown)


projects/gui/src/mainwindow.cpp
```diff
@@ -898,6 +908,19 @@ QString MainWindow::genericTitle(const TabData& gameData) const
...
+{
+	QPointer<ChessPlayer> white = m_players[Chess::Side::White];
+	QPointer<ChessPlayer> black = m_players[Chess::Side::Black];
+	bool isHumanGame =  (!white.isNull() && white->isHuman())
+			 || (!black.isNull() && black->isHuman());
+	bool gameOn = (!m_game.isNull() && !m_game->isFinished());
+	m_adjudicateBlackWinAct->setEnabled(gameOn);
+	m_adjudicateWhiteWinAct->setEnabled(gameOn);
+	m_adjudicateDrawAct->setEnabled(gameOn);
+	m_resignGameAct->setEnabled(gameOn & isHumanGame);
```
ilaripih: I think you meant to use the logical `&&` operator instead of the bitwise `&`?

projects/gui/src/mainwindow.cpp
```diff
@@ -1075,6 +1098,26 @@ void MainWindow::adjudicateGame(Chess::Side winner)
 				  Q_ARG(Chess::Result, result));
 }
 
+void MainWindow::resignGame()
+{
+	if(m_game.isNull() || m_game->isFinished())
```
ilaripih: Coding convention violation: no space between `if` and `(`.

projects/gui/src/mainwindow.cpp
alwey: Indeed, this should be the && operator. Thank you! :-)

alwey: changed, thank you!

PR #344, opened 2018-01-20
Enable translation by auto-detecting system locale.

Prcuvu:
This PR makes GUI load translations by auto-detecting system locale.
It would be better to do this step by step. Previous PR will be closed.


ilaripih:
Looks good, thanks!

Issue #139, opened 2016-09-13
(POSIX) Show full username instead of login name


kervinck:
Don't show login name but the full username.

```
diff --git a/projects/gui/src/cutechessapp.cpp b/projects/gui/src/cutechessapp.cpp
index 83714f8..918600e 100644
--- a/projects/gui/src/cutechessapp.cpp
+++ b/projects/gui/src/cutechessapp.cpp
@@ -23,6 +23,11 @@
 #include <QFileInfo>
 #include <QSettings>

+#ifndef Q_OS_WIN32 // POSIX
+ #include <pwd.h> // for: struct passwd, getpwuid
+ #include <unistd.h> // for: getuid
+#endif
+
 #include <mersenne.h>
 #include <enginemanager.h>
 #include <gamemanager.h>
@@ -103,7 +108,14 @@ QString CuteChessApplication::userName()
        #ifdef Q_OS_WIN32
        return qgetenv("USERNAME");
        #else
-       return qgetenv("USER");
+       struct passwd *passwordEntry = getpwuid(getuid());
+       if (!passwordEntry || !passwordEntry->pw_gecos)
+               return QString::fromUtf8("No name");
+       // Isolate first field from GECOS data in case there are any
+       char *s = strchr(passwordEntry->pw_gecos, ',');
+       if (s != NULL)
+               *s = '\0';
+       return QString::fromUtf8(passwordEntry->pw_gecos);
        #endif
 }

```

kervinck:
Slightly better (to catch the case when the system doesn't have a full name):
```
diff --git a/projects/gui/src/cutechessapp.cpp b/projects/gui/src/cutechessapp.cpp
index 83714f8..e2594f7 100644
--- a/projects/gui/src/cutechessapp.cpp
+++ b/projects/gui/src/cutechessapp.cpp
@@ -23,6 +23,11 @@
 #include <QFileInfo>
 #include <QSettings>
 
+#ifndef Q_OS_WIN32 // POSIX
+ #include <pwd.h> // for: struct passwd, getpwuid
+ #include <unistd.h> // for: getuid
+#endif
+
 #include <mersenne.h>
 #include <enginemanager.h>
 #include <gamemanager.h>
@@ -103,7 +108,17 @@ QString CuteChessApplication::userName()
        #ifdef Q_OS_WIN32
        return qgetenv("USERNAME");
        #else
-       return qgetenv("USER");
+       struct passwd *passwordEntry = getpwuid(getuid());
[...]

Issue #110, opened 2016-09-03
Label players' colors more clearly


elcabesa:
whatching a game it's difficult to understand who is white and black.

ilaripih:
White is always on the left and black on the right. Would it help if it said "White" over the white player's name and "Black" over the black player's name?

elcabesa:
I think it oculd help :)

ilaripih:
I'll have to think about this. Most chess GUIs are just using the "white on the left, black on the right" convention. But I like what PyChess has done with the little white or black analog clock next to the player name.

ilaripih:
Thanks @alwey !

PR #347, opened 2018-01-23
User Name Options

alwey:
This PR suggests adding choices for the name representation of human players to the GUI
- Add option to use full user name instead of login name
- Add option to display sides of players on chess clock
- Checkboxes to configure these options in General Settings

References: Issues #139 and #110 .

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -552,7 +552,9 @@ void MainWindow::setCurrentGame(const TabData& gameData)
 			auto clock = m_gameViewer->chessClock(side);
 			clock->stop();
 			clock->setInfiniteTime(true);
-			clock->setPlayerName(gameData.m_pgn->playerName(side));
+			QString name = nameOnClock(gameData.m_pgn->playerName(side),
+						   side);
+			clock->setPlayerName(name);
 		}
 
 		updateWindowTitle();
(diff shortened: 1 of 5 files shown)


projects/gui/src/mainwindow.cpp
```diff
@@ -898,6 +901,16 @@ QString MainWindow::genericTitle(const TabData& gameData) const
 		       .arg(white, black, result.toShortString());
 }
 
+QString MainWindow::nameOnClock(const QString& name, Chess::Side side) const
+{
+	QString text = name;
+	bool displaySide = QSettings().value("ui/display_players_sides_on_clocks", false)
+				      .toBool();
+	if (displaySide)
+		text.append(tr(" (%1)").arg(side.toString()));
```
ilaripih: `Side::toString()` already returns a localized string. I think we should just use `QString(" (%1)").arg(side.toString())` here.

alwey: Thank you. I did not know that and I have not looked it up in the documentation, sorry. I will change this accordingly.

alwey:
Changed. There is some concern regarding Q_OS_WIN64 and Q_OS_ANDROID.

ilaripih:
I think this is fine. Android is not a priority at the moment so if we break it we can fix it later.

PR #350, opened 2018-01-30
Continue fifty-moves counting when castling.

alwey:
The half-move counter continues counting up for castling moves. I was surprised to hear that. Indeed, FIDE rules 5.2 and 9.3 and the specification of PGN and FEN, 16.1.3.5 mean that in standard chess pawn moves and captures reset the counter, castling moves do not. I checked and Xboard and PyChess indeed count correctly.

This is also relevant for all FIDE derivative variants (derived from `WesternBoard` and having castling). Drop moves are a different matter, FIDE rules do not specify drop moves.

I suggest to insert a comment into WesternBoard and leave out the counter reset for castlings.
This has consequences for the testing suite, so I adapted the FEN test strings.

[Laws of Chess](http://www.fide.com/FIDE/handbook/LawsOfChess.pdf)
[PGN spec](http://www.thechessdrum.net/PGN_Reference.txt)
[Chessprogramming Wiki about irreversible moves](https://chessprogramming.wikispaces.com/Irreversible+Moves)
[Discussion on open-chess.org](http://www.open-chess.org/viewtopic.php?f=3&t=2209)

I hope this is useful.

### DIFF
--- a/projects/lib/src/board/westernboard.cpp
+++ b/projects/lib/src/board/westernboard.cpp
@@ -879,7 +879,8 @@ void WesternBoard::vMakeMove(const Move& move, BoardTransition* transition)
 			Piece rook = Piece(side, Rook);
 			setSquare(rookSource, Piece::NoPiece);
 			setSquare(rookTarget, rook);
-			isReversible = false;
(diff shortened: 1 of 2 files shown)


ilaripih:
Good find, thanks!

PR #353, opened 2018-02-01
Add zh_CN localization.

Prcuvu:
Here comes a translation file for zh_CN localization. It works when the executable is placed like this:
```
├ translations/
│  ├ cutechess_[language_code_1].qm
│  └ cutechess_[language_code_2].qm
└ cutechess-gui
```
To maintainers: please compile with `lrelease` and include the `.qm` file with the `translations` directory in release packages, excluding the `.ts` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,36 @@ seconds:
 See `cutechess-cli -help` for descriptions of the supported options or manuals
 for full documentation.
 
+Translation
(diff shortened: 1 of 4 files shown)


ilaripih:
Thanks a lot, looks good! We should probably do an official release soon.

PR #358, opened 2018-02-09
Add GUI configuration: human player can/cannot play after timeout

alwey:
This adds a checkbox to the General Settings of the GUI. The user may configure whether a human player can play after timeout or forfeit games on time. The configuration is saved and comes into action _for new games only_ (remark: behaviour of updated version changed) immediately when changing the status of the checkbox.
I refrained from using a per-game configuration. Nevertheless, I hope this is useful.

### DIFF
--- a/projects/lib/src/humanplayer.cpp
+++ b/projects/lib/src/humanplayer.cpp
@@ -18,17 +18,11 @@
 #include "humanplayer.h"
 #include "board/board.h"
 
-
 HumanPlayer::HumanPlayer(QObject* parent)
 	: ChessPlayer(parent)
 {
 	setState(Idle);
 	setName("Human");
-}
-
-bool HumanPlayer::canPlayAfterTimeout() const
-{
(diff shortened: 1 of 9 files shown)


projects/lib/src/humanplayer.cpp
```diff
@@ -28,7 +28,8 @@ HumanPlayer::HumanPlayer(QObject* parent)
 
 bool HumanPlayer::canPlayAfterTimeout() const
 {
-	return true;
```
ilaripih: So far we've managed to keep all `QSettings` stuff in GUI. In order to maintain this separation, I think we need a method in `ChessPlayer` for setting the `canPlayAfterTimeout` value and the GUI should call that method.

alwey:
Changed in order to separate QSettings from lib. The configuration is now passed on from `NewGameDialog` via `HumanBuilder` to `ChessPlayer`.

Issue #321, opened 2017-12-06
Lock engine as white for tournament games


Nordlandia:
I miss the option to select engine as white only in tournament tab. 

For example right click on engine in _Player_ tab then «Lock engine as white 🔒»

Another option or possibility is an checkbox named «Alternate colours» which can be enabled or deactivated if wanted.

alwey:
what kind of tournaments would be relevant?

Nordlandia:
Locking engine is particularly aimed for 1v1 matches, so for Round Robin.

alwey:
We already had related enhancements for the CLI ( #95 and #188). What is missing now is a reasonable GUI implementation.

alwey:
@Nordlandia : /wrt to the [question at talkchess](http://talkchess.com/forum/viewtopic.php?topic_view=threads&p=744498&t=66093): It has been in the **CLI** for a couple of months.

The relevant options are `-noswap` and `-repeat [n]`, see documentation.

Giovanni might want to give it a try.

alwey:
I am not sure I would like to see support of these features (no side swap and  repeat n times) in the GUI. This is why I have not started working on this for the GUI just after the CLI. There is some  prospect of users mixing everything up and then complaining. In my opinion the user of the CLI is more conscious of the settings and his choices.

Nordlandia:
Locking engine as white in matches can be good choise if that particular engine is weaker. 

For example lock _stockfish 8_ against latest _stockfish dev_ to see how it performs.

PR #371, opened 2018-03-29
Scrolling of move list to follow the text cursor. Resolves #292.

### LINKED ISSUE #292: Long move list doesn't scroll when user is navigating in game using buttons or keyboard

alwey:
This patch prevents the text cursor from escaping the visible area.

### DIFF
--- a/projects/gui/src/movelist.cpp
+++ b/projects/gui/src/movelist.cpp
@@ -256,6 +256,7 @@ void MoveList::selectChosenMove()
 	m_moves[moveNum].move.mergeCharFormat(c, format);
 
 	c.endEditBlock();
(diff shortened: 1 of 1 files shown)


Nordlandia:
Please consider PR #321

alwey:
@Nordlandia: Hello, yes I am very aware of issue 321. 

But I feel that within an issue or especially within a PR we should try to stay on topic. Recently some users tried to draw attention towards certain aspects of cutechess when posting under unrelated issues. If everybody does this, then we will certainly have lively discussions but more noise and less gain. So I would like to kindly ask for a bit of restraint in order to help the project.

ilaripih:
Nice and short, thanks!

PR #378, opened 2018-04-27
Support Grand Chess

alwey:
- Add support for multi-digit notations
- Add extension to support CECP quirk for games with 10 ranks

This patch supports _Grand Chess_, a 1984 chess variant by Christian Freeling on a 10x10 board. This variant uses piece types known from _Capablanca Chess_. The starting position resembles the position of Embassy Chess. However, both sides have 10 pawns on their _third_ rank, the rooks are at the corners of the board, and the rest of the pieces on their second rank.

grand1

This variant has initial double step option for pawns and en passant captures. Pawns can only be promoted to an own captured piece. Pawn promotion is optional when reaching rank eight or nine but mandatory when reaching the last rank. If promotion is impossible, then the pawn will not be able to advance to the tenth rank. It will be able to give check to a king on the end rank nevertheless.

The implementation uses `GrandBoard`, a subclass of `CapablancaBoard`. The base classes have been adapted to support notation with multiple digits, currently for board sizes up to at least 26x99 (one character, two digits). `Board::moveFromLanString`, `WesternBoard::sanMoveString` and `WesternBoard::moveFromSanString` have been modified aiming at low loss of execution speed for standard chess. 

A peculiarity of the Xboard engine protocol (CECP) is the shift of rank notation from 1-10 to 0-9 for boards with exactly 10 ranks. The method` XboardEngine::transformMove` takes care of that.
`
The implementation has been tested with both UCI and CECP engines Sjaak II 1.4.1 (both protocols) and NebiyuAlien (CECP).

I hope this is useful.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -81,6 +81,8 @@ Fischer Random Chess / Chess 960
 Giveaway Chess (Losing Chess)
 .It gothic
 Gothic Chess
+.It grand
+Grand Chess
 .It grid
 Grid Chess
 .It gridolina
--- a/projects/cli/res/doc/help.txt
+++ b/projects/cli/res/doc/help.txt
@@ -33,6 +33,7 @@ Options:
 			'fischerandom': Fischer Random Chess/Chess 960
 			'giveaway': Giveaway Chess (Losing Chess)
 			'gothic': Gothic Chess
+			'grand': Grand Chess
 			'grid': Grid Chess
 			'gridolina': Berolina Grid Chess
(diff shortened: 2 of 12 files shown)


musketeerchess:
Hi
It is possible to add seirawan chess and musketeer chess?

developed by @IanFab

2018-04-28 0:51 GMT+02:00 alwey <notifications@github.com>:

alwey:
@musketeerchess: Seirawan Chess is already in since January 2018.
Regarding Musketeer Chess, there are issues regarding trademarks and/or patents that would have to be solved first.

PR #383, opened 2018-06-21
Leela times out on startup often

jjoshua2:
Especially with 20 block network cutechess is unreliable to start up leela lc0 without timing out. It takes about 9s on my computer if nothing else is running.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -28,3 +28,8 @@ test_script:
     - nmake /NOLOGO
     - cd parser & release\tst_jsonparser & cd..
     - cd serializer & release\tst_jsonserializer & cd..
+
+artifacts:
+
+  - path: build/Release/cutechess-cli.exe
+    name: cutechess-cli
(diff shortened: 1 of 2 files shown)


jjoshua2:
This actually fixed your appveyor so it uploads now. And it fixes so leela doesn't time out!!

ilaripih:
Cute Chess has a separate timeout of 35 seconds for initializing the chess protocol. Does Leela initialize the network graph and/or weight after the UCI handshake or is there a bug in Cute Chess that effectively caps the initialization time at 10 seconds? Or does Leela need at least 9s between every move?

jjoshua2:
Yes, leela initializes after the UCI handshake I guess. If you load it and type UCI it's instant. But if you type go there's 9s before the first move. Then there is no initialization with other moves. I tested this patch and it fixed the issue. I think it's designed to wait for the isready command to initialize. If it initializes on startup it takes too long for some GUIs to load and check parameters, like chessbase, although I heard they also did a patch to increase timeout values...

ilaripih:
I guess the easiest solution is to merge this one, although I find it strange that other GUIs would allow long delays between moves but only a short delay for initialization.

PR #391, opened 2018-07-05
Support Los Alamos Chess

alwey:
This PR adds support of Los Alamos Chess.

Los Alamos chess is a variant of standard chess which is played on a 6x6 board. This variant has no bishops, no pawn double steps, and no castling. It was introduced in 1956 by Paul Stein, Mark Wells, and
others with the first chess-like program on the MANIAC I computer of Los Alamos Scientific Laboratory. Due to the limited resources of the computer it was not yet practical to implement standard chess.

This implementation has been tested with _Sjaak II 1.4.1_, _NebiyuAlien_, and _Fairy-Stockfish_.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -107,6 +107,8 @@ King of the Hill Chess
 Knightmate
 .It loop
 Loop Chess (Drop Chess Variant)
+.It losalamos
+Los Alamos Chess
 .It losers
 Loser's Chess
 .It makruk
--- a/projects/cli/res/doc/help.txt
+++ b/projects/cli/res/doc/help.txt
@@ -46,6 +46,7 @@ Options:
 			'kingofthehill': King of the Hill Chess
 			'knightmate': Knightmate
 			'loop': Loop Chess (Drop Chess)
+			'losalamos': Los Alamos Chess
 			'losers': Loser's Chess
 			'makruk': Makruk (Thai Chess)
 			'modern': Modern Chess (9x9)
--- a/projects/lib/src/board/board.pri
+++ b/projects/lib/src/board/board.pri
@@ -48,6 +48,7 @@ SOURCES += $$PWD/board.cpp \
     $$PWD/oukboard.cpp \
     $$PWD/aseanboard.cpp \
     $$PWD/sittuyinboard.cpp \
+    $$PWD/losalamosboard.cpp \
     $$PWD/boardfactory.cpp \
     $$PWD/boardtransition.cpp \
     $$PWD/syzygytablebase.cpp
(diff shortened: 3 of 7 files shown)


ianfab:
@alwey Thanks a lot for the recent additions of variants, and also for testing my new Stockfish fork. Have you observed any issues with it, especially for los alamos chess? The implementation for smaller board sizes is still a bit experimental (e.g., some parts of the evaluation are nonsense at the moment, like a passed pawn evaluation that assumes an 8x8 board, but at least should not crash), so although I also tested it a bit myself, I was not sure whether it already is fully stable.

alwey:
@ianfab : You are welcome. I will notify you if I find problems. What I noticed is that `losalamos` and `aiwok` do not support the "go perft n" command yet.

ianfab:
What do you mean by "do not support (perft)"? Does it crash or report wrong numbers? Please note that after changing the variant, you always need to send a `position` command, otherwise it still uses the previous position (and variant).

For `losalamos` I get the same perft numbers on the starting position as http://www.eglebbk.dds.nl/program/chess-perft.html up to depth 8, so it looks fine to me. It does not generate double steps, but I noticed that I forgot to filter out double steps in the validation of TT moves, so it could be that illegal double steps make it into the search tree after some time (which should however not affect perft). Thanks, I'll fix that.

alwey:
I tested with version ee5ac30:

setoption name UCI_Variant value ai-wok
info string variant ai-wok startpos rnsaksnr/8/pppppppp/8/8/PPPPPPPP/8/RNSKASNR w - - 0 1
go perft 5
...
Nodes searched: 4865609

Sjaak II 1.4.1 and cutechess give 13275068 for ai-wok and 4865609 for standard chess (!)

setoption name UCI_Variant value losalamos
info string variant losalamos startpos rnqknr/pppppp/6/6/PPPPPP/RNQKNR w - - 0 1
go perft 5
...
Nodes searched: 4865609

Sjaak II 1.4.1 and cutechess give 191846

ianfab:
@alwey That is exactly what I meant above. You need to send a position command (e.g., `position startpos`) after the setoption UCI_Variant, otherwise it continues to play the previous position (e.g., the standard chess starting position). In principle I think this should be compliant with how the UCI protocol works in general, but I could make it more intuitive by always resetting to the starting position of a newly selected variant.

(Sorry for hijacking the thread...)

alwey:
@ianfab : Oh, my bad. Forgot this, of course it is necessary to use the position command to be standard compliant. Now `losalamos` perft 5: 191846, `ai-wok` perft 5:  13275068. All values match. Sorry for the noise.

ianfab:
Ok, thanks for confirming. The double step issue should also be [fixed](https://github.com/ianfab/Fairy-Stockfish/commit/e0eb3b58dc67109e14f0cd58f05a2bc66ad756e5) now.

alwey:
@ianfab: I will fetch the newest version. Just tested, losalamos perft 8 all match.

PR #392, opened 2018-07-05
[PREVIEW, WIP, HELP NEEDED] Add function to pause a game via GUI menu item

alwey:
This PR aims to add the functionality to pause a game (softly).

If the action "Pause Game" is chosen the side to move should not be affected and should complete their move normally.  The game is paused after the side to move has made their move. After the desired pause the action "Resume Game" should start the opponent side's turn.

Help is welcome especially /wrt designing and testing the functionality.
The combinations of UCI and CECP protocol and pondering engine / non-pondering engine / human player must be considered - and also varying sequences of "Pause Game" and "Resume Game".

Harder measures like interrupting, stopping, aborting games may also be considered in this context.

Remarks:
Introduced `halt` method to ChessGame
Added `isPaused` method to ChessGame
Added `stopPondering` method to ChessEngine, XboardEngine and UciEngine
Invoked `stopPondering` in ChessGame::onMoveMade method
Send stop only if `isPondering()` in `UciEngine::stopPondering()`
Added Pause Game and Resume Game to Game Menu


projects/lib/src/chessgame.cpp
```diff
@@ -296,7 +301,15 @@ void ChessGame::onMoveMade(const Chess::Move& move)
 void ChessGame::startTurn()
 {
 	if (m_paused)
+	{
+		if (!playerToMove()->isHuman())
```
ilaripih: You don't have to check if the player is human; just call `playerToMove()->stopPondering()` like you do in the `halt()` method.

projects/lib/src/chessgame.cpp
```diff
@@ -647,6 +660,16 @@ void ChessGame::pause()
 	m_paused = true;
 }
 
+void ChessGame::halt()
```
ilaripih: Looks like this method is not used anywhere. Is it needed in the future?

projects/lib/src/chessgame.cpp
alwey: Ah right, `ChessPlayer::stopPondering()` saves this. So no need to test for `isHuman()`.

projects/lib/src/chessgame.cpp
ilaripih: Okay. I'd rather avoid committing unused code. But you can always add a TODO comment about planned improvements to the code.

ilaripih:
I tested this briefly. A couple of things popped up:

* Even with infinite time control in human-human games the Pause/Resume functionality is there but it doesn't really do anything.
* In human-human games making a move will in effect resume the game (the move is made, turn changes, clock counts down) but the menu still gives the "Resume" option. When a human makes a move the menu Pause/Resume function should automatically go to the "game in progress" mode where "Pause" is the available action.
* If a human should make the first move in the game both "Pause" and "Resume" should probably be disabled initially because the game will only start when the first move is made. Or is there a use case for explicitly starting the chess clock before the first human move?

In engine-engine games this seems to work well.

alwey:
Thank you for testing. Apart from the problems you listed for human games, I sometimes encountered problems in engine-engine matches (ponder on).

ilaripih:
Oh right, I didn't test with pondering. It can be very tricky, especially on the UCI side.

alwey:
Removed unused `halt()` method from `ChessGame`. Added keyboard shortcuts for "Pause" and "Resume".  Now the `HumanPlayer::wokeUp` signal is also used in order to update menus.

fsmosca:
Tried this in win 10, matching 2 engines with ponder on, one cecp the other is uci.
#### Crafty vs Deuterium
Crafty played Nf3
Crafty is pondering

Deuterium played d5 with ponder move c4
Deuterium is pondering

Crafty is thinking for best move after Nf3 d5

I pause the game while crafty is thinking and Deuterium is pondering.

Crafty sends its best move c4

Cutechess sends stop command to Deuterium to stop its pondering. Correct here is ponderhit but perhaps a stop is just fine since this game is about to be paused. Current position should be saved disregarding the ponder move / ponderhit.

Deuterium sends bestmove e7e6, as reply to stop command.

Game is paused at this moment.

...

Later I press resume to continue the game.

Cutechess sends this to Deuterium:
`>Deuterium v2019.2.37.73(7): go wtime 175152 btime 288681 winc 5000 binc 5000`

### The issue
The problem begins here, because Deuterium did not receive the position command from Cutechess.
From here on, the game is out of sync.

It should have been,
`position startpos moves g1f3 d7d5 c2c4`
c2c4 was the move from Crafty just before the game was paused. So startpos moves g1f3 d7d5 c2c4 is the current position.

followed by,
`go wtime 175152 btime 288681 winc 5000 binc 5000`

projects/lib/src/chessgame.cpp
alwey: Unused code removed

alwey:
@fsmosca : Thank you for testing this and sending this bug report. Highly appreciated. I will try to improve the restart sequence.

alwey:
A recent version with improvements for pondering engines. Thanks to @fsmosca  for the decisive help. 

Older changes: Removed unused `halt()` method from `ChessGame`. Added keyboard shortcuts `P` for "Pause" and `Space` for "Resume".  Now the `HumanPlayer::wokeUp` signal is also used in order to update menus.

> I tested this briefly. A couple of things popped up:

> Even with infinite time control in human-human games the Pause/Resume functionality is there but it doesn't really do anything.

Now the functionality is active. But for infinite time control it does not make a difference and can be ignored.

 >   In human-human games making a move will in effect resume the game (the move is made, turn changes, clock counts down) but the menu still gives the "Resume" option. When a human makes a move the menu Pause/Resume function should automatically go to the "game in progress" mode where "Pause" is the available action.

Now the "game in progress" mode is reached after the player has made their move. Nevertheless for human play the delayed pause style feels a bit awkward.

>    If a human should make the first move in the game both "Pause" and "Resume" should pSpacerobably be disabled initially because the game will only start when the first move is made. Or is there a use case for explicitly starting the chess clock before the first human move?

I believe there is: over the board the chess clock is started by the opponent playing Black.

Issue #285, opened 2017-10-15
Paste PGN/FEN into the main board


andreaspaulick:
This could be useful for engine analysis of a game/position in the computers clipboard

alwey:
+1

This is also related to #284.
Some elements have to be implemented (and to be clarified) to yield this functionality.

Thoughts:
A board in a main window would be a possible target (technically there is no main board now). The action would not influence this board directly, but a new board would be created.
The PGN paste gesture should set the PGN variant (opening a new board dialog and possibly closing the target board later). FEN does not carry a variant, so one would use the variant of the target window.  The PGN or FEN data are regarded as opening sequence with very high depth. 

The type (and possibly the style) of the intended action might be selected by a menu, or e.g. by (not) pressing modifier buttons like SHIFT, CONTROL, etc. .

Some settings have to be taken from configuration or from the target board, e.g. time control, White and Black players, adjudication rules.

PR #395, opened 2018-07-06
Pasting general FENs to targeted variants

alwey:
Small patch: Suggestion to get the variant information from the targeted (template) game.
I tested this with standard chess, crazyhouse and grid chess.

A new game might also "inherit" other properties from the targeted game, like time control settings, engine setups, etc. "Paste & Inherit".

### DIFF
--- a/projects/gui/src/mainwindow.cpp
+++ b/projects/gui/src/mainwindow.cpp
@@ -977,12 +977,17 @@ void MainWindow::pasteFen()
 	if (cb->text().isEmpty())
 		return;
 
-	auto board = Chess::BoardFactory::create("standard");
+	QString variant = m_game.isNull() || m_game->board() == nullptr ?
+				"standard" : m_game->board()->variant();
+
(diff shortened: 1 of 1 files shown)


artoj:
I like the idea. I think the inherit should be separate function and the normal paste should ask for the variant using some method (I've discussed with Ilari but we haven't reached a conclusion yet).

Could you separate this functionality to its own menu item?

alwey:
OK, I will try to come up with an additional function.

alwey:
Related to issue #285

alwey:
The "Paste & Inherit" functionality takes more effort than initially anticipated. So I will open a new PR for that.

I would suggest to use the patch above to amend the normal Paste functionality. For many users the patch will make no difference, playing standard chess only. But for variant players I guess it would be helpful.

ilaripih:
Okay, I'll merge this one in the meantime because it's still an improvement over the current functionality and probably intuitive for most users.

PR #400, opened 2018-07-09
Add function to paste FEN and use the settings of the target.

alwey:
This function takes variant, time controls, players, and stored settings from targeted tab (template game) and creates a new game with a starting FEN from the clipboard.

Hack: To get the template configurations all engine configurations are searched for matching player names. This does not work with ad-hoc player configurations/names, because they have no entry in the list of saved configurations.

 Function as discussed in #395


projects/gui/src/mainwindow.cpp
```diff
@@ -969,40 +975,108 @@ void MainWindow::copyFen()
...
-		msgBox.setInformativeText(cb->text());
-		msgBox.exec();
+	auto game = new ChessGame(board, new PgnGame());
 
+	pgnOk = pgnOk && game->setMoves(pgnGame);
+	if (!pgnOk &&  !board->setFenString(text))
+	{
 		delete board;
-		return;
+		return showPasteErrorDialog(variant, text);
```
ilaripih: The `ChessGame` object takes ownership of `board` so there's no need to delete `board` here. However, it looks like the `game` and its `new PgnGame()` object are leaked. `ChessGame` doesn't take ownership of its `PgnGame` object; `MainWindow` will, but only after `MainWindow::addGame()`.

projects/gui/src/mainwindow.cpp
```diff
@@ -969,40 +975,108 @@ void MainWindow::copyFen()
...
+	}
+	if(!pgnOk)
+		// action for FEN input
+		game->setStartingFen(text);
+
+	if (useDefault)
+		game->setTimeControl(TimeControl("inf"));
+	else
+	{
+		game->setTimeControl(*m_players[0]->timeControl(), Chess::Side::White);
```
ilaripih: Tournament games and their players are deleted automatically when the game ends. So this code line will crash if you run it when the active tab is showing the last game of a finished tournament. This is one reason why I'd like to have copies of `PlayerBuilder` in `MainWindow::tabData`. Or maybe `TournamentPlayer` because it includes time control settings.

projects/gui/src/mainwindow.cpp
```diff
@@ -969,40 +975,108 @@ void MainWindow::copyFen()
...
+			    tr("Invalid FEN or PGN string for the \"%1\" variant:")
+			    .arg(variant),
+			    QMessageBox::Ok, this);
+	msgBox.setInformativeText(text);
+	msgBox.exec();
+
+	return;
+}
+
+void MainWindow::paste()
```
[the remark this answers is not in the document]
alwey: Sure,`pasteDefault` and `pasteCurrent` are more appropriate.

projects/gui/src/mainwindow.cpp
alwey: Ah, thank you. Now deleting game and pgn instead.

alwey: The main path uses MainWindow::addGame via ChessGame::start and SIGNAL initialized().
If ChessGame::startFailed we do not clean up, AFAIK.

projects/gui/src/mainwindow.cpp
alwey: Thank you.These are very valuable hints. 
Now testing whether `m_player[0].isNull()` or `m_player[1].isNull()`. 
Disabling Menu Item _"Paste FEN/PGN and Use Target Settings"_ if `m_game.isNull()` not only to be safe but to avoid inexpected behaviour for users.

PR #401, opened 2018-07-12
Support Pyramid Tournament Type

alwey:
This PR supports a  "pyramid tournament". This is essentially a round-robin tournament where the sequence of pairings is different. Every listed player meets all of their predecessors in the list, in sequence.

Example: Participants 0, 1, 2, 3. 

First, Player 0 will play against nobody - skipped.
Then player 1 will be paired against player 0 (round 1).
After that player 2 will play 0, then play against 1 (round 2).
Finally, player 3 will play 0, 1, and at last 2 (round 3).

The CLI options `-games n` and `-rounds m` and their counterparts in the GUI can be used to further specify the tournament. They work as for round-robin tournaments.

I hope this is useful.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -190,6 +190,8 @@ Round-robin tournament (default)
 First engine plays against the rest
 .It knockout
 Single-elimination tournament
+.It pyramid
+Every engine plays against all of its predecessors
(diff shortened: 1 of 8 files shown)


ilaripih:
Looks good to me, thanks!

Issue #284, opened 2017-10-15
Load a game from the database window into the main programm window (for analysis)


andreaspaulick:
It would be useful if you could load the database games into the main program window for analysing purposes (for example via an engine)

alwey:
+1

PR #407, opened 2018-07-19
Game export to clipboard from game database

alwey:
Export a single PGN game displayed in the database game viewer via a new button "Copy game".
This adds complementary functionality /wrt to #400.

db1

### DIFF
--- a/projects/gui/src/gamedatabasedlg.cpp
+++ b/projects/gui/src/gamedatabasedlg.cpp
@@ -25,6 +25,7 @@
 #include <QModelIndex>
 #include <QFileDialog>
 #include <QInputDialog>
+#include <QClipboard>
 
 #include <pgnstream.h>
 #include <pgngame.h>
@@ -40,6 +41,8 @@
 #include "pgndatabase.h"
 #include "gamedatabasesearchdlg.h"
 #include "threadedtask.h"
+#include "cutechessapp.h"
+#include "board/board.h"
 
 #ifdef QT_DEBUG
 #include <modeltest.h>
@@ -250,6 +253,7 @@ void PgnExportTask::run()
 GameDatabaseDialog::GameDatabaseDialog(GameDatabaseManager* dbManager, QWidget* parent)
 	: QDialog(parent, Qt::Window),
 	  m_gameViewer(nullptr),
+	  m_game(),
 	  m_dbManager(dbManager),
 	  m_pgnDatabaseModel(nullptr),
 	  m_pgnGameEntryModel(nullptr),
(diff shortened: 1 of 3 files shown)


alwey:
Second, incremental commit:
Export a single position displayed in the database game viewer as FEN via a new button "Copy FEN".

db2

alwey:
Relates to #284

projects/gui/src/gamedatabasedlg.cpp
```diff
@@ -302,6 +306,18 @@ GameDatabaseDialog::GameDatabaseDialog(GameDatabaseManager* dbManager, QWidget*
 	connect(ui->m_createOpeningBookBtn, SIGNAL(clicked(bool)), this,
 		SLOT(createOpeningBook()));
 
+	connect(ui->m_copyGameBtn, SIGNAL(pressed()), this, SLOT(copyGame()));
+	connect(ui->m_copyGameBtn, &QPushButton::released, this, [=]()
+	{
+		ui->m_copyGameBtn->setText(tr("Copy game"));
```
ilaripih: Maybe it would be more consistent to use "Copy PGN" and "Copy FEN" or "Copy game" and "Copy position". I like the first option better because then the user will know the format of the game notation and we use "PGN" in a lot of other places too.

projects/gui/src/gamedatabasedlg.cpp
```diff
@@ -527,11 +544,39 @@ void GameDatabaseDialog::createOpeningBook()
 	task->start();
 }
 
+void GameDatabaseDialog::copyGame()
+{
+	QString str("");
```
ilaripih: I think empty strings are usually initialized with the default constructor, so `QString str;` should be enough.

projects/gui/src/gamedatabasedlg.cpp
alwey: changed accordingly

projects/gui/src/gamedatabasedlg.cpp
alwey: Ah, thank you. Have been unsure at that moment.

projects/gui/src/gamedatabasedlg.cpp
```diff
@@ -436,6 +452,7 @@ void GameDatabaseDialog::gameSelectionChanged(const QModelIndex& current,
 	ui->m_resultLabel->setText(game.tagValue("Result"));
 
 	m_gameViewer->setGame(&game);
+	m_game = game;
```
ilaripih: Now that you've created the `m_game` object you don't need `game` anymore and will save a copy operation.

alwey: Ah, thank you. Reusing `m_game` works, because `PgnGame::read()` calls `PgnGame::clear()`. I did not look into the details of PgnGame then. So I used a fresh copy.

PR #409, opened 2018-07-26
Add option to discard unfinished games from PGN output

### LINKED ISSUE #230: option for discarding unfinished games from tournament PGN output
Split from #209 

TournamentSettingsDialog should have a checkbox for discarding unfinished games from PGN output, for example when the tournament is stopped early.

alwey:
This patch adds the option to omit unfinished games from PGN output. When stopping a tournament also stalled and disconnected games are left out.

The CLI has an additional optional parameter for `-pgnout` :
```
-pgnout FILE [min][fi]
                        Save the games to FILE in PGN format. Use the 'min'
                        argument to save in a minimal/compact PGN format. Only
                        finished games are written when argument 'fi' is set.
```
TournamentSettingsDialog of the GUI has an additional checkbox "Save unfinished games".
The default behaviour is unchanged: cutechess GUI and CLI will write all PGN games.
This resolves #230 .

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -1,4 +1,4 @@
-.Dd February 8, 2018
+.Dd July 26, 2018
 .Dt CUTECHESS-CLI 6
 .Os
 .Sh NAME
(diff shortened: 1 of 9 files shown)


ilaripih:
Cool, thanks!

PR #411, opened 2018-07-29
CECP engine forfeits on claim of illegal move

### LINKED ISSUE #109: Forfeit if Xboard engine claims illegal move
Cute Chess is never supposed to send illegal moves to engines. So if an Xboard engine claims that it received an illegal move, we can assume that it's buggy and should forfeit the game.

alwey:
As long as cutechess has an arbiter function, illegal moves are not (read: should not be) transmitted to engines. So a claim of an illegal move should not be valid. At present, without adjudication the claiming engine waits and finally forfeits on time.

This patch avoids the wait and adjudicates against the CECP engine claiming an illegal move. This resolves #109 .

### DIFF
--- a/projects/lib/src/xboardengine.cpp
+++ b/projects/lib/src/xboardengine.cpp
@@ -759,6 +759,13 @@ void XboardEngine::parseLine(const QString& line)
 			pos += rx.matchedLength();
 		}
 	}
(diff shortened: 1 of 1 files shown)


ilaripih:
Assuming that Cute Chess is always right. I like it.

PR #430, opened 2018-10-07
Support Hoppel-Poppel and New Zealand Chess

alwey:
This patch adds support for two leightweight chess variants with hybrid piece movement. These variants are known at least since the 1920s.

Hoppel-Poppel Chess is standard chess but with modified Bishops and Knights. The Bishop captures like a standard Knight but else moves as usual. This piece configuration is called `Biskni`. The Knight captures like a standard Bishop but makes their usual hops when not capturing. This piece is called`Knibis`. In the endgame these hybrid pieces seem to be weaker than Bishop or Knight.

New Zealand Chess is similar. The Bishops are normal Bishops, but here the Rooks and the Knights are hybrids called `Rookni` (R moves and N captures) and `Kniroo` (N moves and R captures). The Rooknis can participate in castling. In the endgame these hybrid pieces are clearly weaker than a Rook and perhaps as useful as a Knight.

`HoppelPoppelBoard` has been tested with [Fairy-Stockfish](https://github.com/ianfab/Fairy-Stockfish/). Starting position perft 6 match.

I hope this is useful.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -103,6 +103,8 @@ Grid Chess
 Berolina Grid Chess
 .It gryphon
 Gryphon Chess
+.It hoppelpoppel
+Hoppel-Poppel (has N/B hybrids)
 .It horde
 Horde Chess (v2)
 .It janus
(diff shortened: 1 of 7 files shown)


alwey:
rebased

PR #442, opened 2018-10-27
Option to use two-sided scores for resign adjudication

### LINKED ISSUE #385: cutechess-cli resign
The game adjudication of the resign option is based on the score of only one engine. It could be useful to adjudicate the game if the losing engine is below a certain score and the winning engine is above that score.

alwey:
This PR resolves #385.

The CLI option `-resign` now has a new optional parameter `twosided`. When set to _true_ it activates two-sided resign adjudication: In addition to the conditions of the normal (single-sided) resign adjudication the winning side's scores must be at least _SCORE_ centipawns above zero for at least _COUNT_ consecutive moves.

So the proposed usage for two-sided adjudication is

`cutechess-cli ... -resign movecount=4  score=650 twosided=true  ...`.

Normal, single-sided resignation works as before.

In the GUI there is a new group of radio buttons to select either normal (single-sided) or two-sided resignation score adjudication.

----
Original text:

The option `-resign2`  is added to the CLI. The new option activates two-sided resign adjudication. It has no parameters of its own. It only has an effect when the `-resign `option is in action: In addition to the conditions of the normal (single-sided) resign adjudication the winning side's scores must be at least _SCORE_ centipawns above zero for at least _COUNT_ consecutive moves.

So the proposed usage is 
`cutechess-cli ... -resign movecount=4  score=650 -resign2 ...`

rather than
`cutechess-cli ... -resign movecount=4  score=650 twosided=true  ...`.

In the GUI there is a new group of radio buttons to select either normal (single-sided) or two-sided resignation score adjudication.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -1,4 +1,4 @@
-.Dd July 26, 2018
+.Dd October 25, 2018
 .Dt CUTECHESS-CLI 6
 .Os
 .Sh NAME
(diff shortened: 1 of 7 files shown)


ilaripih:
Thanks! The code looks fine but I prefer the `twosided` parameter over `-resign2` because it's self-documenting and it's a sub-parameter to `-resign` like it should be. You don't need `twosided=true`; It could be just `twosided`, like the `min` and `fi` sub-parameters of the `-pgnout` parameter.

alwey:
Thank you. The suggested arrangement of parameters actually has been my first preference, too. However I had difficulties with the properties of `MatchParser::Option::toMap` which treats listed arguments as mandatory, not optional.

alwey:
I will try to extend `MatchParser::Option::toMap` so that it also takes a QString of optional arguments.

PR #443, opened 2018-10-30
Add qt511 to travis

rikardfalkeborn:
Also bump to 5.9.6.


alwey:
Why bump higher w/o need?

rikardfalkeborn:
> Why bump higher w/o need?

Hmm, I figured it'd be good to test the latest released version, but perhaps it's more important to test the oldest release available? I can remove it if you prefer it.

alwey:
No, its OK.  I cannot speak /wrt the MS Windows environment.

In linux context some of the most used 5.9 versions seem to be 5.9.1 (Ubuntu 18.04 Release) and 5.9.5 (Ubuntu 18.04 Proposed) and soon perhaps 5.9.6.

However, I guess the most used version of Qt under linux would actually be 5.5.1, which is already out of scope.

ilaripih:
I'd say it's a good rule of thumb that we test the latest stable version of every major release, beginning from the lowest supported version (Qt 5.7).

artoj:
The original idea was to take the Qt version shipped in Debian Stable and use that as the lowest supported version. I think this is a good compromise since the some people running Cute Chess may prefer running Stable so we want to support them too.

Currently Debian Stable ships 5.7.1. When Stable updates ~H2 2019, we'll bump the minimum requirement again though probably not to 5.11 since some Ubuntu LTS releases are shipping older versions of Qt.

PR #445, opened 2018-11-04
Evaluation history for finished tournament games.

### LINKED ISSUE #233: tournament games' evaluation history is gone when game ends
This is yet another side effect of deleting the `ChessGame` object after a tournament game has ended: when the active tab is switched to the finished game, the evaluation history can't be retrieved. We could store that info in the `PgnGame` object, and in fact it is already stored there in the form of PGN comments.

alwey:
Uses PGN comments to reconstruct the evaluation history.

Resolves #233


ilaripih:
Heh, you actually did it (parsing the history from PGN comments). I'll merge this because it's an effective hack to fix the bug. But I'll add a comment to the source to remind me to get back to this when I refactor GUI game management a bit.

PR #457, opened 2018-11-12
Rounding Elo to one decimal place.

zz4032:
Reason for this suggestion is that it's very rare that Elo estimations have error bars low enough to justify a precision of two decimal places.


alwey:
That is true. It might also be true for

- tournaments with more than two players
- error margin field
- LOS

PR #480, opened 2019-01-12
Fix output for CLOP script

gonzalezjo:
It seems that there is a *severe* error in the current CLOP script. Wins and losses are swapped, meaning that the script is optimizing for losing games and avoiding wins(!). This pull request fixes that.

This is a documented issue. See here for more details: https://groups.google.com/d/msg/fishcooking/0uRw0Zila0M/s2siUjVzFFwJ


gonzalezjo:
...false alarm. :innocent:

PR #483, opened 2019-01-14
Add id number of engine issuing an illegal PV move, and print PV line

### LINKED ISSUE #471: Add ID to name in Warning message
a cutechess-cli 'wish-for-improvement' 

In the following output (as obtained with -debug)
```
2395864 <cluster(405): info depth 21 seldepth 31 multipv 1 score cp 66 nodes 11674383 nps 91924275 tbhits 0 time 127 pv h7g6 g3f3 c4d6 d5b6 e7g5 b6d5 g5e7 f3f4 d6c4 e2e1 e7d6 f4g5 c4e5 g5f6 e5d3 e1e2 f8e8 e6e7 d3c5 f6g7 f5e6 c3c4 d6e7 d5c7 e8d7 c7e6
[...]
Warning: Illegal PV move g3f3 from cluster
```
would be nice if the name displayed in the warning would be cluster(405) instead of cluster, so it is somewhat easier to find the details.

alwey:
This small patch causes output of some more information in case of illegal PV moves.
Resolves #471

### DIFF
--- a/projects/lib/src/uciengine.cpp
+++ b/projects/lib/src/uciengine.cpp
@@ -826,9 +826,14 @@ QString UciEngine::sanPv(const QVarLengthArray<QStringRef>& tokens)
 		auto move = board->moveFromString(token.toString());
 		if (move.isNull())
 		{
-			qWarning("Illegal PV move %s from %s",
-				 qUtf8Printable(token.toString()),
-				 qUtf8Printable(name()));
+			QString tokenString(token.toString());
+			qWarning("Illegal PV move %s from %s (%d)",
+				 qUtf8Printable(tokenString),
+				 qUtf8Printable(name()),
+				 id());
+			qWarning("PV: %s %s",
+				 qUtf8Printable(pv),
+				 qUtf8Printable(tokenString));
 			break;
 		}
 		if (!pv.isEmpty())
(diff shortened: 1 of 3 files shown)


projects/lib/src/uciengine.cpp
```diff
@@ -826,9 +826,14 @@ QString UciEngine::sanPv(const QVarLengthArray<QStringRef>& tokens)
 		auto move = board->moveFromString(token.toString());
 		if (move.isNull())
 		{
-			qWarning("Illegal PV move %s from %s",
-				 qUtf8Printable(token.toString()),
-				 qUtf8Printable(name()));
+			QString tokenString(token.toString());
+			qWarning("Illegal PV move %s from %s (%s)",
```
ilaripih: I think you can use `%d` instead of `%s` and `id()` instead of `qUtf8Printable(QString::number(id()))`

alwey: Very true. I did not think about formats other than %s. Changed accordingly.

alwey:
updated

ilaripih:
Sorry it took so long to back to Cute Chess, we've been very busy. Merging this pull request will mark the beginning of the 2019 Cute Chess hackathon!

Issue #461, opened 2018-12-01
7-man syzygy adjudication


Nordlandia:
On talkchess thread, syzygy author comfirmed WDL is suffice for simple adjudication. However cutechess idoesn't automaically adjudicate at times if 7-man wdl files is at SSD drive. Is it cutechess related or few files missing like checksum.md5 and 3-4-5-6.md5 from 6-man syzygy set.

Ipmanchess:
We have still a waiting 6-men adjudication..still doesn't work..

0zymandia2:
6 men has worked for me since they implemented syzygy adjudication. I don't know about 7 men.

Ipmanchess:
Then you have them all in one folder these WDL & DTZ files.. i keep them seperate and you have no adjudication..

0zymandia2:
In the command line, try something like "option.SyzygyPath=C:\dtz;D:\wdl"

SyzygyPath is the most common engine option, some might have a different nomenclature.

Ipmanchess:
Check here issue 370: https://github.com/cutechess/cutechess/issues/370

0zymandia2:
I was talking command line, I've never used the GUI.

Nordlandia:
CuteChess won't adjudicate 6-man positions for some reasons. 

SSD: 960 EVO - 6-Man WDL placed on that M.2 Drive.

Issue #370, opened 2018-03-26
Tablebase Adjudication


Ipmanchess:
I would better make a seperate Ipman requests folder ;)

Next i tried to find out is why i can't use Tablebase Adjudication?
On my laptop i use 5-men syzygy ,windows 7 and no problem to select Tablebase Adjudication.

On my new system i use 6-men syzygy ,windows 10 , till Jerehmia on TCEC chat told me ,it's your directory pointing to two different folders as my 6-men syzygy has a dtz & wdl folder ..so i use C:\RTB6\dtz**;**C:\RTB6\wdl
but when closing settings and re-open ,Cutechess cutt-off the second part..and Tablebase adjudication stays light grey and can't select it? It keeps C:\RTB6\dtz without the wdl part..

So is their a way to point my tablebases right.. or can this be added..i like to keep these syzygy TB's clean and it would be a pitty to put them all together in one folder..
Again with FritzGui is this no problem..so it was normal for me to use same way in Cutechess..seperate directory name/folder but with ";" or "," doesn't work..

So one directory works -> C:\RTB  and can select Tablebase Adjudication!
Using two folders not -> C:\RTB6\dtz;C:\RTB6\wdl

Nordlandia:
I store whole 6-men inside one folder on an external HDD. The HDD wakes up from idle mode once 6-pieces is present on the board. Probing 5-men + 6-men adjudication is good alternative to probing 6-men.

Nordlandia:
Wdl is sufficient for simple adjudication. Tested and confirmed by using PCIe SSD.

jjoshua2:
I also want the GUI to be able to specify at how many pieces the adjudication should be performed at. Especially with partial 7 piece syzygy out now, I might only want adjudication at 4 or 6 pieces to see more endgame.

Nordlandia:
@jjoshua2 Yes, that's right.

Ipmanchess:
I did a test with **100games using TC: 5m+5s** on a i9 7980XE @4.2Ghz using 6-men syzygy
Stockfish - asmFish : averange depth's during games : 37.80 / 37.24
With FritzGui it takes **22u14m** ,and with Cutechess Gui to finish these 100games it takes **36u15m**
It runs 14hours longer because i see these endless draw games playing till end..
This fix would save us a lot time!!

Is there any plan or date when we gonne see a new release version..
Thanks,Ipman.

PR #486, opened 2019-01-20
Wdl not dtz should be the minimum required for adjudication.

Tilps:
While in fathom source code probe_root calls probe_dtz, probe_dtz actually ends up calling probe_ab, which calls probe_wdl_table.
Without WDL tables this fails, which cascades upwards to probe_root failing.

WDL files should be the expected minimum - dtz files for syzygy are technically not complete without the corresponding WDL files.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -201,7 +201,7 @@ equals zero (default).
 Adjudicate games using Syzygy tablebases.
 .Ar Paths
 should be semicolon-delimited list of paths to the compressed tablebase files.
-Only the DTZ tablebase files are required.
+Only the WDL tablebase files are required.
 .It Fl tbpieces Ar N
 Only use tablebase adjudication for positions with
 .Ar N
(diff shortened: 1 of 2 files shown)


Nordlandia:
I clearly remember that previously only wdl was suffice for simple static adjudication. Now in cutechess i need full syzygy set for adjudcation. I have them stored on HDD because i have much space left. For analysis i have WDL on 250GB NVMe M.2. 
See #461

Tilps:
The code is a bit strange - I could see possibility of needing both wdl and dtz to get the adjudication code to work in its current form.  But definitely if adjudication occurs only at 50 move reset points syzygy should be able to adjudicate using just WDL.  Once you are past that point DTZ can help you adjudicate, but even then only if there has been no repetitions.
DTZ on its own is definitely not enough though.  So the documentation should change, even if perhaps not as per this PR>

ilaripih:
We should definitely implement issue #370 later, but I'll merge this now.

PR #498, opened 2019-04-05
Improvements of SittuyinBoard

alwey:
This PR only affects SittuyinBoard and consists of three patches:

1. Implementation of Sittuyin rule 3.9c.7 (Maung Maung Lwin: How to play Myanmar Traditional Chess, p. 8): If promotion moves are the only legal moves to avoid a stalemate then the weaker side may claim a draw. So cutechess will adjudicate a draw with this software patch.

sittuyin1
In this situation white to move may claim a draw. FEN: 1k4PK/3r4/8/8/8/8/8/8[-] w - 0 0 1

2. There is a problem with pawn promotions in engine play. This patch avoids wrong decoding of promotion moves in LAN format (by then throwing away the result of moveFromSanString for promotion moves in SittuyinBoard!).

3. A problem in `SittuyinBoard::isLegalPosition` gets obvious when using a debug compile and running the chessboard test suite: The move history must not be empty when using `Board::lastMove()`.

### DIFF
--- a/projects/lib/src/board/sittuyinboard.cpp
+++ b/projects/lib/src/board/sittuyinboard.cpp
@@ -195,6 +195,17 @@ bool SittuyinBoard::vSetFenString(const QStringList& fen)
 	return ret;
 }
 
+Move SittuyinBoard::moveFromSanString(const QString& str)
+{
+	Move move = MakrukBoard::moveFromSanString(str);
+
+	// Avoid problems with promotion moves in LAN(!) format
+	if (move.sourceSquare() != 0
+	&&  move.promotion() != Piece::NoPiece && !str.contains("=") )
+		return Move();
+
+	return move;
+}
 
 void SittuyinBoard::vMakeMove(const Move& move, BoardTransition* transition)
 {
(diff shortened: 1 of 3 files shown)


projects/lib/src/board/sittuyinboard.cpp
```diff
@@ -301,6 +327,12 @@ Result SittuyinBoard::result()
 			return Result(Result::Draw, Side::NoSide, str);
 		}
 	}
+	if (!m_inSetUp && !canMakeNormalMove())
```
ianfab: @alwey I did not test it, but from the code it reads like a position such as `8/8/4p3/4P3/2p5/2P3p1/5k2/7K[-] w - - 0 1` would also be adjudicated as a draw, although 3.9c7 seems to explicitly restrict the adjudication to cases where only one pawn is left:
> When no other pieces but only a pawn is left on the board and it has no square to move and forming the stale-mate position, the player has the right not to take promotion and claim the draw.

So I guess something like the below should need to be added.
```suggestion
	if (!m_inSetUp && pieceCount(sideToMove(), Pawn) == 1 && !canMakeNormalMove())
```

alwey: @ianfab : thank you for the heads up. Indeed the wording of rule 3.9c7 says "only one Pawn".
Other positions where additional Pawns are in opposition then do not match.

alwey: I added your suggestion and also a fix for a bug in my latest patch.

PR #504, opened 2019-04-29
Introduce policies when to pick the next opening in tournaments

### LINKED ISSUE #501: Opening not identical for all gauntlet opponents
When using option `-tournament gauntlet` with engine1 playing against engine2 and engine3, the games are played with opening1 against engine2 and a different opening2 against engine3 in the same round.
Options `-repeat` and `order=sequential` have no influence on that issue.

alwey:
Resolves #501
(library and CLI).

Alternative 2 - "More general solution"

This patch adds a parameter _policy_ to the CLI option `-openings`,e.g. `policy=default`.
It can have one of three values:

_default_: Use a new opening when a new encounter (pair of players) starts or the specified number of repetitions is reached. Cutechess behaves as before - except for some cases of incommensurable combinations n,m of -repeat n and -games m.

_encounter_  uses a new opening only for a new encounter.

_round_ shifts to a new opening only when a new round starts.

The _RoundPolicy_ currently depends on the (tag) round counters of the tournament types. These are not in sync with the schedule repetition counters controlled by the -round n option of the CLI or the corresponding `m_roundsSpin` SpinBox of the GUI.

Reference: PR #502

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -1,4 +1,4 @@
-.Dd October 25, 2018
+.Dd April 29, 2019
 .Dt CUTECHESS-CLI 6
 .Os
 .Sh NAME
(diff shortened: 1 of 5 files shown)


ilaripih:
I merged the other solution but now that I think about it I actually I like this general solution better.

PR #506, opened 2019-05-18
Extend repetition settings in the GUI

alwey:
Up to now in the GUI each opening (from a repertoire of PGNs or EPDs ) was only played either once or twice in a tournament, then the next opening was selected. This PR extends the choice. This has already been available in the CLI (request #188, PR #269). 

In the `TournamentSettingsWidget` the checkbox _"Play each opening twice"_ is replaced by a spinbox _"Play each opening n times"_. For the number `n` it offers all choices that match (are factors of) the number of games per encounter. For _gauntlet tournaments_ there are additional choices (multiples of the number of games per encounter that also match the number of rounds to be played).

I hope this is useful.

rep1

tourn1

### DIFF
--- a/projects/gui/src/gamerepetitionspinbox.cpp
+++ b/projects/gui/src/gamerepetitionspinbox.cpp
@@ -49,11 +49,11 @@ void GameRepetitionSpinBox::setTournamentType(const QString& tournamentType)
 // Returns the highest matching number of games
 int GameRepetitionSpinBox::limit() const
 {
-	int games = m_gamesPerEncounter;
+	int games = qMax(m_gamesPerEncounter, 1);
 	if (m_tournamentType != "gauntlet")
 		return games;
 
-	int factor = m_rounds;
+	int factor = qMin(m_rounds, 10000000 / games);
 	while (m_rounds % factor)
 	{
 		factor--;
--- a/projects/gui/src/gamerepetitionspinbox.h
+++ b/projects/gui/src/gamerepetitionspinbox.h
@@ -63,7 +63,6 @@ class GameRepetitionSpinBox : public QSpinBox
 		int limit() const;
 		QValidator::State examine(int value) const;
 
-		int m_maxGames;
 		int m_gamesPerEncounter;
 		int m_rounds;
 		QString m_tournamentType;


projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
+	{
+		static int store = 1;
+		int value = valueIn;
+		int step = (ui->m_repeatSpin->value() > store) ? 1 : -1;
+		int games = ui->m_gamesPerEncounterSpin->value();
+		int rounds = 1;
+		if (ui->m_gauntletRadio->isChecked())
+			rounds = ui->m_roundsSpin->value();
+
+		while (games % value
```
ilaripih: This is an infinite loop when `games == 1` and `rounds == 1` (ie. the default state when someone opens the Settings dialog)

projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
+		ui->m_roundsSpin->valueChanged(value);
+	});
+
+	// repeats dependent on games per encounter, rounds and tournament type
+	connect(ui->m_repeatSpin,static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged),
+		[=](int valueIn)
+	{
+		static int store = 1;
+		int value = valueIn;
+		int step = (ui->m_repeatSpin->value() > store) ? 1 : -1;
```
ilaripih: Shouldn't `ui->m_repeatSpin->value()` be the same as `valueIn` here so we could just use that?

projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
+	{
+		int value = ui->m_roundsSpin->value();
+		ui->m_roundsSpin->valueChanged(value);
+	});
+
+	// repeats dependent on games per encounter, rounds and tournament type
+	connect(ui->m_repeatSpin,static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged),
+		[=](int valueIn)
+	{
+		static int store = 1;
```
ilaripih: I don't like this global variable here; a member variable would be better.

projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
+		limit = (limit > 999999 ) ? 999999 : limit;
+		ui->m_repeatSpin->setMaximum(limit);
+	});
+
+	// update repeats after rounds
+	connect(ui->m_roundsSpin,
+		static_cast<void (QSpinBox::*)()>(&QAbstractSpinBox::editingFinished), [=]()
+	{
+		int value = ui->m_repeatSpin->value();
+		ui->m_repeatSpin->valueChanged(value);
```
ilaripih: It's okay to call slots directly but I think signals should always be used with the `emit signal(param);` syntax.

projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
+		int games = ui->m_gamesPerEncounterSpin->value();
+		int rounds = 1;
+		if (ui->m_gauntletRadio->isChecked())
+			rounds = ui->m_roundsSpin->value();
+
+		while (games % value
+		&&     (value % games || (games * rounds) % value) )
+			value += step;
+
+		ui->m_repeatSpin->setValue(value);
```
ilaripih: I thinks it's quite questionable to call `setValue` from `valueChanged`; there's a risk of infinite loop.

projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
+		store = ui->m_repeatSpin->value();
+
+		// set focus after automatic adjustment
+		if ( valueIn != value )
```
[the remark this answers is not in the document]
alwey: fixed

projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
 	});
 
+	// number of rounds limits repeats
+	connect(ui->m_roundsSpin, static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged),
+		[=](int value)
+	{
+		if (!ui->m_gauntletRadio->isChecked())
+			value = 1;
+		int limit = value * ui->m_gamesPerEncounterSpin->value();
+		limit = (limit > 999999 ) ? 999999 : limit;
```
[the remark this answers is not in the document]
alwey: Of course, fixed.

projects/gui/src/tournamentsettingswidget.cpp
```diff
@@ -32,6 +32,73 @@ TournamentSettingsWidget::TournamentSettingsWidget(QWidget *parent)
...
+		ui->m_repeatSpin->setMaximum(limit);
+	});
+
+	// number of games per encounter limits repeats
+	connect(ui->m_gamesPerEncounterSpin,
+		static_cast<void (QSpinBox::*)(int)>(&QSpinBox::valueChanged), [=](int value)
+	{	int rounds = ui->m_gauntletRadio->isChecked() ?  ui->m_roundsSpin->value() : 1;
+		int limit = value * rounds;
+		limit = (limit > 999999 ) ? 999999 : limit;
```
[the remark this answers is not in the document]
alwey: Fixed.

projects/gui/src/tournamentsettingswidget.cpp
alwey: Of course, a member variable is better: fixed.

projects/gui/src/tournamentsettingswidget.cpp
alwey: Changed accordingly.

projects/gui/src/tournamentsettingswidget.cpp
alwey: Indeed, verified. Changed accordingly.

projects/gui/src/tournamentsettingswidget.cpp
alwey: This is indeed a recursion, but the call depth will not exceed 2.

projects/gui/src/tournamentsettingswidget.cpp
alwey: When `games==1` and `rounds ==1` we have the condition

`(1 % value && ( value % 1 || ( 1 * 1 ) % value) )`

This is `( 1 % value && ( false || 1 % value) )` , or simply `(1 % value)`

Because `value` will always start between 1 and `rounds * games` (the maximum is controlled by the values the `m_gamesPerEncounterSpin` and `m_roundsSpin` SpinBoxes) and under the condition of the while loop it cannot leave this range, so no infinite loop occurs.

projects/gui/src/tournamentsettingswidget.cpp
alwey: I have just simplified this, so the setter is only called if needed.

projects/gui/src/tournamentsettingswidget.cpp
alwey: An infinite loop may indeed occur for gauntlet tournaments, because `rounds * games` may overrun. Protection needed.

projects/gui/src/tournamentsettingswidget.cpp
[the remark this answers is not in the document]
alwey: retired after simplification

projects/gui/src/tournamentsettingswidget.cpp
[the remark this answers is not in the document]
alwey: Retired after simplification.

projects/gui/src/tournamentsettingswidget.cpp
alwey: I restructured the logic in this method and removed two other connect methods.

alwey:
I will also try out the suggested solution via `m_repeatSpin::editingFinished`.

alwey:
I made some of the suggested modifications: Now there is a `GameRepetitionSpinBox` derived from `QSpinBox`.  The value update uses `GameRepetitionSpinBox::validate()`, `GameRepetitionSpinBox::fixup()`, and `GameRepetitionSpinBox::stepBy()`. When editing the QWidgets `m_gamesPerEncounterSpin`, `m_roundsSpin` and `m_gauntletRadio` is finished then` GameRepetitionSpinBox::interpretText` is called. This updates the value of the `GameRepetitionSpinBox`. I encountered no infinite loops with this version.

projects/gui/src/tournamentsettingswidget.cpp
alwey: Reworked.

ilaripih:
I fixed a couple of bugs, removed some hard dependencies and pushed the changes to master. @alwey can you test if the spin boxes work as expected? If they do, this PR can be closed.

A really good feature BTW!

alwey:
Testing ...

ilaripih:
Thanks for testing! Now it should work for gauntlet as well.

alwey:
Thank you. It works. I like your solution very much.

I would suggest using a fixed value, say 10,000,000, to limit the possible number of games (for gauntlet).

Initialising as `m_maxGames(10000000)`, and testing this bound in limit()
`while ((long)games * (long)factor > m_maxGames || m_rounds % factor)`

As an alternative `m_maxGames` may be deleted.

Currently one can enter really big numbers for both  games per encounter and number of rounds so that the GUI may take very long to calculate the next step.

ilaripih:
I did delete `m_maxGames` and didn't notice any slowness even with extreme values of `m_rounds` and `m_gamesPerEncounter`.
But the m_repeatSpin value did overflow to negative so we'll have add some hard upper bound on the repetition count.

alwey:
I made a change to limit the repetitions to 10 millions in order to speed up and avoid overflows.

ilaripih:
> I made a change to limit the repetitions to 10 millions in order to speed up and avoid overflows.

Nice! I just noticed that the current while loop (without the repetition limit) did nothing but waste CPU cycles.

ilaripih:
I merged this manually because I wanted to change the latest commit's description.

ilaripih:
BTW, in C++17 there's an std function for finding the greatest common divisor: https://en.cppreference.com/w/cpp/numeric/gcd
Too bad we can't use it yet.

PR #508, opened 2019-06-03
Have additional cutechess.exe artifact

jjoshua2:
CLI and non CLI are both useful. Might take a few tries to get syntax accepted.


jjoshua2:
Tested working now

ilaripih:
Thanks! Seems that Arto just implemented this.

PR #525, opened 2019-09-28
Fix problem reading EPD openings

alwey:
Small fix because EPD opening books do not work anymore in the GUI. 

(Qt::CaseInsensitive has value 0.)


artoj:
Good find. I'll see if I can create a LGTM query of this. I'll commit the query first and then this.

artoj:
Merged, thanks! I added a LGTM query so this type bugs should not happen again.

PR #526, opened 2019-09-28
Copy (and modify) configurations in settings list of engines (GUI)

alwey:
This PR adds a method to copy ("clone") an engine configuration from the list of configured engines (Tools->Settings->Engines). This makes it easier to use modified engine settings under a dedicated name. Initially the copied settings have the name of the original configuration prepended by "_From". This can be renamed and the settings can be modified immediately.  I hope this is useful.

clone0

clone1


artoj:
Committed, thanks!

PR #527, opened 2019-09-28
End the SettingsGroup in TimeControl::writeSettings

alwey:
This is a small fix. The SettingsGroup in writeSettings has no termination.


artoj:
Committed, thanks!

Issue #387, opened 2018-06-27
GUI: Powerfull Features Request


d3vv:
1. Hot key for Flip(rotate) Board.

This often needed when two people play on the same pc. It's comfortable against boring mouse actions which steal a game time.

2. Long time (hours, days, etc.) play with engine.

On most engines we can't play in infinite mode due to we are infinite wait engine's move. Need time limit option (+ponder) for engine move when play against human. IMHO this is an honest measure.

3. Analyze Mode(window)/Mate finder + "Compute move" hotkey from specific engine.

This really needed for Engine developers.

4. Hotkey for Copy/Paste FEN + Position Setup.

Need for developers and others.

5. Hotkey for move "take back" action.

General feature for education.

alwey:
Hello, thank you for your support. Your list comprises of 5 feature requests. Usually the developers prefer to separate each feature into distinct issues in order to keep track of these features independently.

> 
>     Hot key for Flip(rotate) Board.
> 
> This often needed when two people play on the same pc. It's comfortable against boring mouse actions which steal a game time.

Indeed, using keys is faster.
Please try pressing Alt-V and then F.

> 
>     Long time (hours, days, etc.) play with engine.
> 
> On most engines we can't play in infinite mode due to we are infinite wait engine's move. Need time limit option (+ponder) for engine move when play against human. IMHO this is an honest measure.

I think this is already possible to configure long time controls via the "New Game" and "New Tournament" dialogs. However it is not possible yet to configure per player time controls in the GUI. This has already been listed as issue #161 and referenced in issue #351.

Ponder is available as "thinking on opponents time" in the GUI. In the CLI one can use ponder as an engine option.

> 
>     Analyze Mode(window)/Mate finder + "Compute move" hotkey from specific engine.
> 
> This really needed for Engine developers.

Indeed, currently there is no Analyze Mode: **New Feature**.


> 
>     Hotkey for Copy/Paste FEN + Position Setup.
> 
> Need for developers and others.

FEN: Please try pressing Alt-G and then E, then paste via Paste-Key (usually Ctr-V) into target.
PGN: Please try pressing Alt-G and then P, then paste via Paste-Key into target.

> 
>     Hotkey for move "take back" action.
> 
> General feature for education.

Currently, there is no "Take Back" action, because cutechess's home turf is Engine-Engine matches. This has been requested, please refer to #287.

ilaripih:
Maybe we should automatically flip the board after every move in human-human games?

[...]

PR #530, opened 2019-10-05
New shortcut for board flipping

alwey:
This small patch for the GUI assigns shortcut _CTRL-F_ to flipping the view of the board.

Related to #387


artoj:
Is Ctrl+F a common shortcut for this action? If yes, ok @artoj

artoj:
It seems that each GUI has their own keyboard shortcut for this feature, for example F2, F12 and Ctrl+R. I think Ctrl+F is fine and we can change it later it we want more consistency between other GUIs.

PR #531, opened 2019-10-06
Fix issue where UCI_Variant is sent without a value.

MortenLohne:
This PR is a bugfix for UCI code specific to the `UCI_Variant` option. The bug happens with any engine that supports `UCI_Variant` combo option with the `standard` value. When playing standard chess matches through cutechess-cli, cutechess will send `setoption name UCI_Variant` to the engine without an actual value. However, in the UCI specification, only button-type options can be set without a value. Doing it for combo-type options is a bug, and causes some engines to crash.

### DIFF
--- a/projects/lib/src/uciengine.cpp
+++ b/projects/lib/src/uciengine.cpp
@@ -54,7 +54,7 @@ QString variantFromUci(QString str, bool uciPrefix = true)
 
 QString variantToUci(const QString& str, bool uciPrefix = true)
 {
-	if (str.isEmpty())
+	if (str.isEmpty() || (str == "standard" && uciPrefix == true))
 		return QString();
 
 	if (str == "fischerandom")
--- a/projects/lib/src/uciengine.cpp
+++ b/projects/lib/src/uciengine.cpp
@@ -54,7 +54,7 @@ QString variantFromUci(QString str, bool uciPrefix = true)
 
 QString variantToUci(const QString& str, bool uciPrefix = true)
 {
-	if (str.isEmpty() || str == "standard")
+	if (str.isEmpty())
 		return QString();
 
 	if (str == "fischerandom")


projects/lib/src/uciengine.cpp
```diff
@@ -54,7 +54,7 @@ QString variantFromUci(QString str, bool uciPrefix = true)
 
 QString variantToUci(const QString& str, bool uciPrefix = true)
 {
-	if (str.isEmpty() || str == "standard")
+	if (str.isEmpty())
```
alwey: if (str.isEmpty() || (str == "standard" && uciPrefix == true))  // to be safe

alwey:
Thanks for the heads-up!

I can confirm there is a problem sending only `setoption name UCI_Variant` and omitting`value standard`.

projects/lib/src/uciengine.cpp
MortenLohne: Is it necessary though? I don't want to add more spaghetti to this code if it isn't needed. That is no fault of cutechess of course, chess variants in uci is a mess and cutechess is just doing its best keep up.

To be honest I don't see why this function should ever return an empty string at all, it just seems like begging for trouble.

alwey: AFAIK it is needed. There is no option `UCI_Standard`. Omitting the second expression sends `setoption UCI_Standard value true` to either engines without `UCI_Variant` combobox option  like mainline _Stockfish_ and _Komodo_   or such engines like _Sjaak II_ with `UCI_Variant` option but with `var chess` instead of`var standard` in the list of values.

MortenLohne: Sorry for forgetting about this PR. I've amended the change like you suggested, and it looks like it is still compatible with latest master

PR #535, opened 2019-10-19
Avoid returning void expressions (get rid of some potential clazy warnings)

alwey:
Clean up a bit.


artoj:
Do you still get Clazy warnings even though those warnings have been excluded with the comment?

alwey:
No, but but to get rid of _potential_ warnings.

artoj:
So I added a couple of LGTM queries to the repo and the one that is relevant to this PR is the one that tries to find returns of void expressions. I don't know whether you can access the LGTM query console and paste the query I committed but there are couple more cases where a similar construct exists. There are already flagged with `clazy:exclude` so you should be able to find them easily. If you could update the PR with those also.

alwey:
Done. I left out the case of clazy:exclude=qstring-arg in enginemanager.cpp:112.

artoj:
Merged, thanks!

Issue #478, opened 2019-01-08
Feature request: Enable file/directory selection for string type in Advanced tab in cutechess-gui


Prcuvu:
For UCI options such as `SyzygyPath`, one may need to specify a path. It is annoying to copy and paste the path from file explorer or terminal. Please consider adding a right-click popup menu containing "Choose file..." and "Choose directory...", which redirects to corresponding file open dialog, to any "string" type value text boxes in the "Advanced" tab of Configure Engine dialog.

alwey:
Actually, most UCI options are engine specific. Speaking of `SyzygyPath`, this can be configured in the GUI once and for all engines via _Tools->Settings-General->"Syzygy tablebases path"_, using the file browser. Is this insufficient?

Prcuvu:
There may be other path-related options that cutechess may not have covered, such as `NalimovPath` (as specified by UCI standard), `GaviotaTbPath`, `Scorpio Egbb Path` and some custom book path. It still stands in demand when encountering such engines.

alwey:
Indeed, `NalimovPath` is specified in the UCI standard. The problem is that in UCI there is no option type "path" to determine that a file path is to be configured. CECP has a path type specification. Perhaps it would be acceptable to infer the specification of a path from the UCI option's name ending on "Path" without being identical to "Path". Then offer selecting paths of the file system. This would not interfere with the UCI spec because this woud only be a convenience hack within the GUI.

ilaripih:
Done

PR #537, opened 2019-10-25
Enable file selector for string type options if name ends with "file"or "path"

alwey:
Enable file selector for string type options if their name ends with "file".
Enable directory selector for string type options if their name ends with "path".

This is a convenience patch.
Resolves feature request #478


artoj:
++Review queue for June hackathon.

ilaripih:
Thanks! Since Xboard engines already have a way to set the correct option type and UCI engines use PascalCase for their option names (eg. "SyzygyPath") I dropped the toLower() part.

PR #539, opened 2019-10-27
Support Shogi (Japanese Chess)

alwey:
This PR supports Shogi, a Japanese form of chess where captured pieces can be brought back ("dropped") into the game, It is played on a 9x9 board. The piece set traditionally consists of
flat five-sided pieces with their names written on them, but this implementation uses the standard piece set of CuteChess.

**Rules:**

Shōgi has eight types of original pieces. Either side starts with a set of 9 x Pawn (on the third rank), 2 x Lance, 2 x Knight, 2 x Silver (General), 2 x Gold (General), 1 x Bishop, 1 x Rook, 1 x King.
The latter three types move like in standard chess. The Pawn moves and captures 1 square straight forward. The Lance moves and captures like a Rook, but only forwards. Similarly, the Knight leaps like a chess Knight, but only in the two forward directions. The Silver goes one square diagonally or forward (5 moves). The Gold goes like a King but not diagonally backward (6 moves).

All these pieces except King and Gold General have the option to promote if their move starts or ends in the opponent's third of the board. A Pawn or Lance reaching the highest rank must promote (otherwise it could not move anymore).  A Knight must promote when reaching the highest two ranks for the same reason. All promoted Pawns, Lances, Knights, and Silvers move like Gold Generals. A promoted Bishop moves like Bishop or King, a promoted Rook like Rook or King.

A captured piece switches sides and goes into the hand of the capturing side. Promoted pieces change back to their original types when captured. A player may drop a piece in hand onto an empty square of the board. This counts as a move. Pawns, Lances, and Knights must not be dropped onto the highest ranks where they would be unable to move. Pawns must not be dropped onto files with own Pawns and must not be dropped to give mate immediately.
 
A side must avert an attack on their King (check) immediately. It is forbidden to enter check. If there is no legal move to avert check then the side in check loses by checkmate.

When a game position occurs four times with the same player to move and the same pieces in hand for each player the game is drawn by repetition. If the situation has arisen from a perpetual check then the side giving check loses.

Rules: http://en.wikipedia.org/wiki/Shogi
 
**Implementation:**

`ShogiBoard` is derived directly from `Board`.  The differences between Standard Chess and Shogi are such that most of the virtual methods had to be overwritten when deriving from `WesternBoard`. or `CrazyhouseBoard`. And then there are additional problems with different notations. So it was better to copy  `WesternBoard` and modify it into `ShogiBoard`. The code length was about the same but flexibility was gained.

Like _Xboard_/_Winboard_ this implementation uses the "normal" coordinate system, chess move notation, western position notation (FEN) and supports  CECP or UCI protocol. Originally in Shogi, the first player (Sente) is Black, the second player (Gote) is White. However, `ShogiBoard` uses the chess assignments: first player White and second player Black.

Tests werde done manually and with CECP engines _Sjaak II_, _CrazyWa_, and _Shokidoki_, and with UCI engines _Sjaak II_ and _Fairy-Stockfish_.

----
If found useful, this might be extended (not in this PR) by
- [ ] United Shogi Interface (USI) for shogi engines
   -  [ ] Support inverted coordinate system 1a...9i
   -  [ ] Notation, especially of promotion and drops
   -  [ ] SFEN (Shogi FEN)
- [ ] Import of Western Shogi notation
- [ ] Import of Japanese notation
- [ ] Speed up

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -145,6 +145,8 @@ Rifle Chess
 S-Chess (Seirawan Chess)
 .It shatranj
 Shatranj
+.It shogi
+Shogi
 .It shoot
 Shoot Chess
 .It simplifiedgryphon
--- a/projects/cli/res/doc/help.txt
+++ b/projects/cli/res/doc/help.txt
@@ -65,6 +65,7 @@ Options:
 			'rifle': Rifle Chess
 			'seirawan': S-Chess (Seirawan Chess)
 			'shatranj': Shatranj
+			'shogi' : Shogi
 			'shoot': Shoot Chess
 			'slippedgrid': Slipped Grid Chess
 			'simplifiedgryphon': Simplified Gryphon Chess
--- a/projects/lib/src/board/board.pri
+++ b/projects/lib/src/board/board.pri
@@ -51,6 +51,7 @@ SOURCES += $$PWD/board.cpp \
     $$PWD/aseanboard.cpp \
     $$PWD/aiwokboard.cpp \
     $$PWD/sittuyinboard.cpp \
+    $$PWD/shogiboard.cpp \
     $$PWD/rifleboard.cpp \
     $$PWD/losalamosboard.cpp \
     $$PWD/almostboard.cpp \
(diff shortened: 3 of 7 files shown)


artoj:
Nice! Ping @ilaripih

alwey:
Thank you!

artoj:
Sorry for the delay. I'll make sure Ilari will review this in the next hackathon next month.

PR #544, opened 2019-11-23
Fix CECP mate scores for winning side

alwey:
This patch fixes a bug I introduced last year into the CECP (xboard) mate score calculation. Only _new style_ mate scores for the winning side are affected. Currently the transformation of the mate scores for the winning side is off one ply e.g. 999998 (+M2) instead of 999999 (+M1).


artoj:
In review queue for June hackathon.

PR #546, opened 2019-11-25
Read halfmove clock and fullmove number from epd opening files

MortenLohne:
This PR fixes an issue where halfmove clocks and fullmove numbers from epd opening files are ignored.

A position in epd format doesn't store halfmove clock and fullmove number in the fen part, but they can still be specified (optionally) through the hmvc and fmvn opcodes. Use this when converting epd to fen.

As a side effect, if hmvc and fmvn are not set, we now return a fen with halfmove clock and fullmove number set to 0 and 1, since returning a fen without those is not strictly correct.

I've checked as carefully as I can that there are no assumptions anywhere that EpdRecord::fen() doesn't return a full fen, so hopefully this change doesn't break anything. As mentioned, the new behavior is more correct anyway.

### DIFF
--- a/projects/lib/src/epdrecord.cpp
+++ b/projects/lib/src/epdrecord.cpp
@@ -44,8 +44,7 @@ bool EpdRecord::parse(QTextStream& stream)
 			return false;
 
 		m_fen.append(tmp);
-		if (i < 3)
-			m_fen.append(" ");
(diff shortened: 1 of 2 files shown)


artoj:
Thanks for the PR! I'll review this with Ilari in the next hackathon in June.

PR #551, opened 2019-12-25
Fix and generalize the book start option.

### LINKED ISSUE #550: cutechess-cli picking openings with start=N
This issues is related to several observations, relating to how the start=N option works, or could work in `-openings file=short.pgn format=pgn order=sequential plies=16 start=6`

The cutechess version used is the one from fishtest, and reports as :
```
~/chess/match$ ./cutechess-cli --version
cutechess-cli 1.1.0
Using Qt version 5.10.1
``` 
Maybe some of these issues are already fixed in master.

* if start=N uses an N larger than the number of openings in short.pgn, the opening book will, seemingly, be ignored, all games start from startpos. Ideally, this option uses wrap-around semantics, like the games actually played.

* if the format is epd instead of pgn, start=N seems to be ignored, i.e. independent of N all games appear to start from the first position in the book.

*Ideally, start=N would also work for order=random. Skipping the first N-1 positions in the random sequence (again with wraparound semantics). This would be the most useful feature to have.

To clarify the use case, fishtest runs batches of games with a given length Nbatch. Right now, the usage of order=random (with a different seed for each batch) leads to repeated games (just by random repetition) in different batches. If instead we would use the same random seed for each batch, but specify start=Nbatch * batchId, we would have different games everywhere (as long as we play less games than the book is long), sample the opening book game randomly, and if we play more games than the book is long, have minimal repetitions. See also the issue:
https://github.com/glinscott/fishtest/issues/478

### LINKED ISSUE #552: cutechess-cli picking openings with random is not fully random
The algorithm used to generate a random permutation of the opening book, is not actually fully random, but the order is biased (games near the beginning of the book are less likely than other book entries to end up near the beginning of the permuted book). 

This leads to measurable effects https://github.com/official-stockfish/Stockfish/issues/2283#issuecomment-568748167

I expect to have a patch soon.

vondele:
Fixes #550 and Fixes #552

* allow the start option to be larger than the number of positions in the book file, using wrapping semantics. Makes it easier to use (no need to know the size of the book, or to restart after N games where N is larger than the size of the book)

* fix the start option to work with epd opening books, right now it was ignored/overwritten.

* introduce the start option to work not only with sequential but also with random order, picking the appropriate location in the random stream.

* fix the generation of the random permutation of the book.

tested to behave as expected looking at a couple of corner cases (start=N, with N=size of the book, size + 1, size + N, etc)

Will also help fixing https://github.com/glinscott/fishtest/issues/478

### DIFF
--- a/projects/lib/src/openingsuite.cpp
+++ b/projects/lib/src/openingsuite.cpp
@@ -19,6 +19,7 @@
 #include "openingsuite.h"
 #include <QFile>
 #include <QTextStream>
+#include <algorithm>
 #include "pgnstream.h"
 #include "epdrecord.h"
 #include "mersenne.h"
@@ -123,7 +124,7 @@ bool OpeningSuite::initialize()
 
 	if (m_order == RandomOrder)
 	{
-		// Create a shuffled vector of file positions
+		// Create a vector of file positions
 		for (;;)
 		{
 			FilePosition pos;
@@ -135,14 +136,14 @@ bool OpeningSuite::initialize()
 			if (pos.pos == -1)
 				break;
 
-			int i = Mersenne::random() % (m_filePositions.size() + 1);
-			if (i == m_filePositions.size())
-				m_filePositions.append(pos);
-			else
-			{
-				m_filePositions.append(m_filePositions.at(i));
-				m_filePositions[i] = pos;
-			}
+			m_filePositions.append(pos);
+		}
+
+		// use a Knuth shuffle to generate a random permutation
+                for (int i = 0; i <= m_filePositions.size() - 2; i++)
+		{
+			int j = i + Mersenne::random() % (m_filePositions.size() - i);
+			std::swap(m_filePositions[i], m_filePositions[j]);
 		}
 
 		if (m_startIndex > m_filePositions.size())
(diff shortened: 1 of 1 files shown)


Chess13234:
@vondele can we just fork it and merge everything fishtest/stockfish needs?
Last commit to master was Oct 27, 2019.

vondele:
while we could maintain a branch, this is overhead, and having things upstreamed is always preferable. 

@artoj any chance you could have a look at this ?

projects/lib/src/openingsuite.cpp
```diff
@@ -129,37 +136,48 @@ bool OpeningSuite::initialize()
...
-			}
+			m_filePositions.append(pos);
+		}
+
+		// use a Knuth shuffle to generate a random permutation
+                for (int i = 0; i <= m_filePositions.size() - 2; i++)
+		{
+			int j = i + Mersenne::random() % (m_filePositions.size() - i);
+			std::swap(m_filePositions[i], m_filePositions[j]);
 		}
```
Sopel97: why not something like
```
auto rng = [](){return Mersenne::random();};
std::shuffle(m_filePositions.begin(), m_filePositions.end(), rng);
```

vondele: The honest answer is that I wasn't aware. Yet, I don't think this is suitable for us as cppreference mentions that: ```Note that the implementation is not dictated by the standard, so even if you use exactly the same RandomFunc or URBG you may get different results with different standard library implementations. ``` while we would like the same random order on cutechess-cli being compiled in different environments.

artoj:
Thanks for the patch. I'll review this with Ilari in the next hackathon next month.

PR #554, opened 2020-01-15
syzygy adjudications: add 7man and 5x speedup

wakamex:
switched to latest tbprobe library [Fathom](https://github.com/jdart1/Fathom) care of @jdart1.

switched cutechess to query using a different API call `tb_probe_wdl` which is the thread safe function for recommended for engines as per the Fathom readme. engines may care about DTZ, but for adjudication only WDL is necessary.


artoj:
The Windows CI failed with this patch. Can you review the build output and submit an update. Thanks!

jdart1:
I think you need to add /Tp to the Windows flags to force compilation of tbprobe.c as a C++ file (Fathom requires compilation under C99 or C++).

ilaripih:
> I think you need to add /Tp to the Windows flags to force compilation of tbprobe.c as a C++ file (Fathom requires compilation under C99 or C++).

I tried that and now I get a bunch of warnings about unsafe functions and an error:
`3rdparty\fathom\src\tbprobe.c(305): error C2664: 'HANDLE CreateFileW(LPCWSTR,DWORD,DWORD,LPSECURITY_ATTRIBUTES,DWORD,DWORD,HANDLE)': cannot convert argument 1 from 'char *' to 'LPCWSTR'`

https://ci.appveyor.com/project/artoj/cutechess/builds/33803803

jdart1:
See https://github.com/wakamex/cutechess/pull/1

PR #561, opened 2020-04-19
Support Knight-Relay Chess

ddugovic:


artoj:
We'll review this in the next hackathon next month.

gaintpd:
> We'll review this in the next hackathon next month.
Could you give some brief introduction to he hackathon activity. As I implement the Xiangqi, minixiangqi, manchu variant. I want to open a pull request, but there exist a boring bug in the minixiangqi implementation that always cause the crash on Windows, but works quite fine on Ubuntu 18.04 LTS, this boring bug prevent me to do so. If the code review acitivy can help me to remove the bug. I think its time to open a pull request right now.

artoj:
@gaintpd Yes, we can try to review it if you create a pull request

gaintpd:
@artoj I have open a pull request to add the support of xiangqi to cutechess

projects/lib/src/board/board.h
```diff
@@ -325,8 +325,12 @@ class LIB_EXPORT Board
 				  const QString& symbol,
 				  unsigned movement = 0,
 				  const QString & gsymbol = QString());
+		/*! Returns true if \a piece on \a square can capture like \a movement. */
+		bool pieceHasCaptureMovement(Piece piece, int square, unsigned movement) const;
+		/*! Returns true if \a piece on \a square can move like \a movement. */
+		bool pieceHasMovement(Piece piece, int square, unsigned movement) const;
 		/*! Returns true if \a pieceType can move like \a movement. */
-		bool pieceHasMovement(int pieceType, unsigned movement) const;
+		bool pieceTypeHasMovement(int pieceType, unsigned movement) const;
```
ilaripih, line 333: Is it intentional that this is not a virtual method (it is reimplemented in derived classes)?

projects/lib/src/board/board.h
```diff
@@ -601,7 +605,17 @@ inline const Move& Board::lastMove() const
...
 
-inline bool Board::pieceHasMovement(int pieceType, unsigned movement) const
+inline bool Board::pieceHasCaptureMovement(Piece piece, int square, unsigned movement) const
+{
+    return Board::pieceHasMovement(piece.type(), square, movement);
+}
+
+inline bool Board::pieceHasMovement(Piece piece, int square, unsigned movement) const
+{
+    return Board::pieceTypeHasMovement(piece.type(), movement);
```
ilaripih, line 615: Maybe add `Q_UNUSED(square)` here.

projects/lib/src/board/board.h
ddugovic: This was unintentional.  I think you are correct that adding the `virtual` keyword would make sense here (I don't recall semantic differences between virtual and pure virtual functions, but I think you are correct that `virtual` is semantically appropriate here).

projects/lib/src/board/board.h
alwey: There are more classes with unused arguments and the new `ShogiBoard`(s).. You might leave out the `square` identifier and only pass the formal argument `int` instead of `int square`.

projects/lib/src/board/board.h
ilaripih: It's actually very important to use virtual functions instead of overridden (non-virtual) functions. We're using pointers to `Chess::Board` pretty much everywhere, and without the `virtual` keyword any `board->pieceHasMovement()` calls would go to `Chess::Board::pieceHasMovement()` when the intention is to call the reimplemented method.

projects/lib/src/board/board.h
```diff
@@ -325,8 +325,12 @@ class LIB_EXPORT Board
 				  const QString& symbol,
 				  unsigned movement = 0,
 				  const QString & gsymbol = QString());
+		/*! Returns true if \a piece on \a square can capture like \a movement. */
+		bool pieceHasCaptureMovement(Piece piece, int square, unsigned movement) const;
```
ilaripih, line 329: This should also be virtual like it is in `KnightRelayBoard`

projects/lib/src/board/board.h
```diff
@@ -325,8 +325,12 @@ class LIB_EXPORT Board
...
+		/*! Returns true if \a piece on \a square can move like \a movement. */
+		bool pieceHasMovement(Piece piece, int square, unsigned movement) const;
```
ilaripih, line 331: This should also be virtual like it is in `KnightRelayBoard`

ilaripih:
A couple of the reimplemented methods are still non-virtual. Also, building ShogiBoard is broken at the moment.

ddugovic:
Right, adding the virtual keywords caused the build failures.  I'm not sure how to resolve that.

ddugovic:
Thanks, I think I have fixed these omissions and errors.

gaintpd:
I have found a way to avoid the modification of the westernboard and board to add the KnightRelay variant support in cutechess.

gaintpd:
I have upload my code to realize the knight relayed pawn move generation. Other pieces can be handled in a similiar way, not implemented yet. The code is now in the knightrelay branch in my fork.
the address is: https://github.com/gaintpd/cutechess/tree/knightrelay

gaintpd:
Remember, my code is based on ddugovic, not from the scratch. Otherwise, It is easily to be regarded that I am a cheater. I have realize the knight-relayed move generation for other pieces. but the complete rule is still out of my awareness.

alwey:
@gaintpd  @ddugovic : I also have an implementation avoiding the modification of both `WesternBoard` and `Board` https://github.com/alwey/cutechess/tree/knightrelay1.

It has been staying on my hard drive for some time. I started coding in order to avoid changing the base classes. But in retrospect I found Daniel's solution more elegant because the code is more compact in total. I would only recommend renaming `defendedByKnight` to `attackedByKnight` or even extend the method to a generic `attackedByPiece`, taking one more argument, the piece type.

alwey:
@gaintpd : You might add knight movement for other pieces than pawns when protected by a friendly knight. Also, the check detection is important when implementing variants with new movements.

gaintpd:
> @gaintpd : You might add knight movement for other pieces than pawns when protected by a friendly knight. Also, the check detection is important when implementing variants with new movements.

Thank you very much for your tips.

gaintpd:
@alwey @ddugovic I think both of you can have a cooperation to make the _Knight-relay_ implementation more graceful and avoid the flutter of the **WesternBoard** and **Board** class.

ddugovic:
I defer to the maintainers about whether avoiding editing code is a goal; I don't maintain this codebase so my opinion is less important than theirs.

ilaripih:
Looking good now. I may still try to move the `defendedByKnight` method to the `KnightRelayBoard` class as it's not used elsewhere.

As for the question about editing code - yes, it's a goal, but if editing code results in significantly less total code than just adding new code then editing is okay.

PR #562, opened 2020-04-19
Update copyright year

ddugovic:


artoj:
Thanks for providing this patch. I think the copyright years on files themselves should updated only if there are major changes in the files and there's no need to update them every year.

Issue #273, opened 2017-08-27
Supporting hourglass time control


Nordlandia:
From wikipedia:

Each player's clock starts with a specified time (e.g. 1 minute, 10 min etc.). While Player 1 is deciding a move, player 1's clock time is decreasing and Player 2's clock time is increasing. This is similar to how an hourglass works; sand empties from one container and fills into the other. Moving slowly gives your opponent extra time. The sum of both clocks will always remain the same. There is no maximum amount of time allotted for a game with this timing method; as long as both players play quickly, the game will continue until its natural end. When time runs out on one player's clock the game is over and that player loses. Use of this time control is uncommon.

Hourglass TC might turn out to be interesting, as i would imagine.

Nordlandia:
Hourglass is available at Shredder website. 

https://www.shredderchess.net/

ilaripih:
Hourglass time control is now available in cutechess-cli. We'll update the implementation when hourglass is supported on the chess protocol level.

Nordlandia:
I appreciate that.

alwey:
It is possible to use, e.g. tc=hg120 (seconds) for the first and tc=hg180 (seconds) for the second chess engine in the CLI, so that the total available time is 300 seconds. Games are played without problems. However, the engines will not allocate optimal thinking times for the hourglass mode, because they do not know about the hourglass specifics. There will be more time available than anticipated when having the next move. But having less time than anticipated would be a much bigger problem.

@ilaripih  @Nordlandia: Now that the basic functionality is accepted, I will try to come up with a patch for the GUI so that human players can try out the hourglass mode.

ilaripih:
Thanks for the GUI patch @alwey, it's merged now.

Nordlandia:
One possible tweak/enhancement to the clock display panel when using hourglass is to let the time get updated concurrently for both parties. I.e. each participant clock get updated every second to keep track of time.

My initial first impression of the clock display is that it's somewhat inconvenient to have to do the calculation in the head to know how much time the opponent get and vise versa. If the clock panel can be updated to sync time, i.e. time goes up and down simultaneously it will be much easier to follow the accumulated/consumed time.

alwey:
@Nordlandia I am checking whether this is easy.

PR #567, opened 2020-05-26
Support hourglass time control

alwey:
This PR adds support for an hourglass-style time control. In this mode the sum of time left for both sides is constant. The time a player used for a move is added to the opponent's available time.

If the time control is specified starting with the `hg` token then the hourglass-mode will be activated.

E.g.: `tc=hg360` means that a side has 360 seconds left at start. If the opponent uses tc=hg180 then there will be 540 seconds available for both sides in total.

Hourglass mode cannot be used with increments. It is active only for both sides else none.

There is no GUI support for hourglass mode yet.

Refers to #273.

### DIFF
--- a/projects/cli/src/main.cpp
+++ b/projects/cli/src/main.cpp
@@ -154,6 +154,7 @@ bool parseEngine(const QStringList& args, EngineData& data)
 			}
 
 			data.tc.setInfinity(tc.isInfinite());
+			data.tc.setHourglass(tc.isHourglass());
 			data.tc.setTimePerTc(tc.timePerTc());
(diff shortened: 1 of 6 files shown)


Nordlandia:
There is a certain chance that the next Shredder program will include hourglass mode. I was in contact with the creator and he said he'll consider it for the next version. The means that the UCI protocol needs to be adjusted. Steffan do if he agrees to it. Unless UCI support it, no engines can understand the format.

ilaripih:
Thank you, looks good! Merged manually (resolved a merge conflict), closing this PR.

PR #575, opened 2020-06-18
Sittuyin: Do not allow promotions to attack an opponent piece directly.

alwey:
This fix implements missing Sittuyin rule 3.9 c.4.

Thanks to @teakado, @gbtami and @ianfab for pointing this out.
Ref.: https://github.com/ianfab/Fairy-Stockfish/issues/14


artoj:
Commited, thanks!

Issue #564, opened 2020-05-12
Can you support Korean chess (janggi)?


cjssh1002:
Can you support Korean chess (janggi)?
I am using the Fairy Stockfish janggi engine. There are a lot of bugs because WinBoard / XBoard doesn't support Korean chess. Korea Janggi has many customers. Can you apply?

gaintpd:
Could you offer me some detailed documents about the rule of Janggi, English is preferred. I have implemented xiangqi support for cutechess. I want to have a try if I can get access to the janggi specification as I live in China where the internet is under the strict censorship that makes me not be able to surf on the internet freely, I can't get the desired information as the guys where the GFW does not exist.

cjssh1002:
dear. 
 
To introduce you the janggi rules, I link to the website address. 
Please re-mail if you need more information. 
And I want to use your discord or messenger for communicate more.
 
https://www.pychess.org/variant/janggi 
http://www.janggi.pl/en/ 
https://en.wikipedia.org/wiki/Janggi#General 
 
PS. Korean chess has special rules that are special from the rules of movement.  
"bikjanggi" (when the king and the king face see each other) It draws or adjudicate the win or loss by the score of pieces.(material count)
And player can use pass. Generally, player can't use same 3 movement repeated.
 
Thank you for your interest in Korean janggi. 
 
 
-----Original Message-----

cjssh1002:
If you have implemented Chinese chess, Korean chess will be very easy. Use the same 10X9 largeboard. Since Winboard does not implement Korean chess, so we use Chinese chess board. If you give me Windows 64 compile, I'll test it.

gaintpd:
Hi cjssh1002， thank you very much for your info and suggestion. I will have a try this weekend.

gaintpd:
Hi cjssh1002, I have finish the implementation of the Janggi support for cutechess. but There may be some rules in Janggi that is missed in the current code. I think the complete functional test should be previous to  the creation of the pr to the cutechess repo.
Here is the screenshot of the Janggi variant in cutechess. the only appearance is different from the normally Janggi that all the pieces is the same size in current implementation whereas the latter one in in different size according the the importance in the game.
![Screenshot from 2020-05-31 21-14-48](https://user-images.githubusercontent.com/17960782/83353248-caff6580-a383-11ea-97f2-fb1daee21ad3.png)

cjssh1002:
Thank you for implementing Korean janggi.

[...]

PR #577, opened 2020-06-27
add janggi(korean chess) support

gaintpd:
There are two features still missing in the current implementation
1. bikjang. The current code is failed to give the right move indication for the Human vs Human and Human vs Engine game.
2.different size piece display on janggi board. This is still missing as this feature needs the skill to modify the janggi pieces one by one.

But the current code is sufficient for the engine developer and engine vs engine games.

Based on the above description. I open this PR to receive a broad opportunity to receive bugfix and enhancement.


gaintpd:
[https://github.com/cutechess/cutechess/issues/564](url)

gaintpd:
Here I want to thank ianfab, a wonderful chess programmer who create a masterpiece of Fairy-Stockfish, a wonderful chess engine which support xiangqi, janggi, minixiangqi, and manchu, without this wonderful engine, I can not bring the minixiangqi, manchu into reality, without this wonderful engine, the janggi implementation can not be the current status. Furthermore,  I get the excellent explanation of janggi rules during the janggi implementation from him.

At last thanks him again, for help me to bring the janggi implementation to reality.

PR #578, opened 2020-06-28
GUI: Add hourglass time control

alwey:
This patch adds initial support of the new hourglass time control to the GUI.
A new radio button is added to the time control dialog.

Reference: Feature request #273

This implementation does not add a textual or graphical indicator to the clock.
Such an indicator may be useful for extensions like hourglass mode or paused games.

This implementation is independent of a possible protocol update for engines.
It should work for existing engines and human players.

I hope this is useful.


Nordlandia:
I can't seem to find the hourglass radio icon in recent compiled cutechess gui with total filesize of  21856 kb.

PR #586, opened 2020-07-05
Support Connect Board Games

alwey:
This patch supports abstract strategy games with the goal of getting n pieces in a line to win - either vertically, horizontally, or diagonally. Thus supports `cfour` (means connect four, game with gravity) and `tictactoe` (game without gravity).

There are chess engines supporting these games. While not playing chess, certain aspects of chess engines can be examined, e.g. /wrt speed of communication or concurrency overhead.

I hope this is useful.

c4-1a


alwey:
On another note, the is a new Gomuku GUI out that is based on CuteChess:  https://github.com/nkg114mc/cutegomoku

PR #588, opened 2020-07-08
Draw Wall squares

### LINKED ISSUE #467: Gustav III's Chess
Variant proposal. 

Chess variant advocated by  King Gustav III of Sweden (1746-92).

>  Strategy: It seems like this game works very well, despite the fact that the combined piece power is enormous compared with regular chess. On this comparably small board, which was also an invention by Gustav III, the General-Adjutant cannot run amok. The positions are too cramped for that. This circumstance explains why the game retains much of the strategical and positional character of regular chess. The General-Adjutant is unique because it can, without any aid, give mate to the enemy king. The four extra squares are also important hiding nests for the King. If a Pawn is worth 1 point, a Queen is worth 9 points, then a General-Adjutant is worth about 12 points. 

gustaviii

http://mlwi.magix.net/bg/gustaviii.htm

alwey:
This patch draws wall squares in a dedicated color.
It aims to improve the scene for Gustav III's chess. Resolves #467.

gustav3

### DIFF
--- a/projects/gui/src/boardview/graphicsboard.cpp
+++ b/projects/gui/src/boardview/graphicsboard.cpp
@@ -61,6 +61,7 @@ GraphicsBoard::GraphicsBoard(int files,
 	  m_coordSize(squareSize / 2.0),
 	  m_lightColor(QColor(0xff, 0xce, 0x9e)),
 	  m_darkColor(QColor(0xd1, 0x8b, 0x47)),
+	  m_wallColor(QColor(0xee,0xee,0xee)),
 	  m_squares(files * ranks),
 	  m_highlightAnim(nullptr),
 	  m_flipped(false)
@@ -108,7 +109,14 @@ void GraphicsBoard::paint(QPainter* painter,
 		rect.moveLeft(rLeft);
 		for (int x = 0; x < m_files; x++)
 		{
-			if ((x % 2) == (y % 2))
+			// referenced board coordinates
+			int file = m_flipped ? m_files - 1 - x : x;
+			int rank = m_flipped ? y : m_ranks - 1 - y;
+			Chess::Square sq(file, rank);
+
+			if (pieceTypeAt(sq).isWall())
+				painter->fillRect(rect, m_wallColor);
+			else if ((x % 2) == (y % 2))
 				painter->fillRect(rect, m_lightColor);
 			else
 				painter->fillRect(rect, m_darkColor);
--- a/projects/gui/src/boardview/boardscene.cpp
+++ b/projects/gui/src/boardview/boardscene.cpp
@@ -418,7 +418,7 @@ GraphicsPiece* BoardScene::createPiece(const Chess::Piece& piece)
 	Q_ASSERT(m_renderer != nullptr);
 	Q_ASSERT(m_squares != nullptr);
 
-	if (!piece.isValid())
+	if (!piece.isValid() && !piece.isWall())
 		return nullptr;
 
 	return new GraphicsPiece(piece,
--- a/projects/gui/src/boardview/graphicsboard.h
+++ b/projects/gui/src/boardview/graphicsboard.h
@@ -144,6 +144,7 @@ class GraphicsBoard : public QGraphicsItem
 		QRectF m_rect;
 		QColor m_lightColor;
 		QColor m_darkColor;
+		QColor m_wallColor;
 		QColor m_textColor;
 		QVector<GraphicsPiece*> m_squares;
 		QPropertyAnimation* m_highlightAnim;


gaintpd:
Hello alwey, there exist a bug in this PR, the paint of the wall square cant handle the case of flipped board, and above all, the sequence of paint of the wall square is not in consist of the fen string process code in board.cpp. the correct code in the graphicsboard should be as below:
`int m = m_flipped ? m_files - x - 1 : x;
 int n = m_flipped ? y : m_ranks - y - 1;
 Chess::Square sq(m,n);`

I have test my modification to it. It can paint the right wall square.

gaintpd:
This bug is found because of I am implementing a new variant that have the asymmetric wall square distribution on the board, I have to add the code of this pr to support the feature of painting wall square.

projects/gui/src/boardview/graphicsboard.cpp
```diff
@@ -108,7 +109,10 @@ void GraphicsBoard::paint(QPainter* painter,
 		rect.moveLeft(rLeft);
 		for (int x = 0; x < m_files; x++)
 		{
-			if ((x % 2) == (y % 2))
+			Chess::Square sq(x,y);
```
gaintpd, line 112: 'int m = m_flipped ? m_files - x - 1 : x;'
'int n = m_flipped ? y : m_ranks - y - 1;'
'Chess::Square sq(m,n);'

'Chess::Square sq(x,y);' should be replaced by the above three line of code

alwey: Thanks for the heads-up. I just tested with an non-symmetrical setup.

alwey: I uploaded a corrected version.

projects/gui/src/boardview/graphicsboard.cpp
```diff
@@ -108,7 +109,14 @@ void GraphicsBoard::paint(QPainter* painter,
 		rect.moveLeft(rLeft);
 		for (int x = 0; x < m_files; x++)
 		{
-			if ((x % 2) == (y % 2))
+			// referenced coordinates
+			int xorig = m_flipped ? m_files - 1 - x : x;
+			int yorig = m_flipped ? m_ranks - 1 - y : y;
```
gaintpd, line 114: Hello alwey. Here the line of code `int yorig = m_flipped ? m_ranks - 1 -y : y` is in a reversed order as it is not in consistent with the fen string processing code in the board.cpp.

If you test the current code with a fen string that the wall squares are distributed asymmetric both in x-axis and y-axis. you will notice the difference. And the right order of the line of code should be
`int yorig = m_flipped ? y : m_ranks - 1 - y;`

gaintpd:
Hello alwey. Now the current code can handle the flipped board. 
But the line of code int 
`yorig = m_flipped ? m_ranks - 1 -y : y `
is in a reversed order as it is not in consistent with the fen string processing code in the board.cpp.

If you test the current code with a fen string that the wall squares are distributed asymmetric both in x-axis and y-axis and flip the board on purpose. you will notice the difference and verify what I said above. And the right order of the line of code should be
`int yorig = m_flipped ? y : m_ranks - 1 - y;`

alwey:
This time I thin I am right. I tested with an asymmetrical set-up.
For variant `gustav3` I used
`arnbqkbnra/*pppppppp*/*8*/*8*/*8*/*9/*PPPPPPPP*/ARNBQKBNRA w KQkq - 0 1`.

The code maps the coordinates back to a rotated by 180 degrees position (xorig, yorig) for flipped boards.

projects/gui/src/boardview/graphicsboard.cpp
alwey: I think the implementation is OK now, please see below.

gaintpd:
Hello alwey, please pay attention to that your fen string and your screenshot is not in consistent with each other. The correspond fen string for your screenshot should be 
`arnbqkbnra/*pppppppp*/*9/*8*/*8*/*8*/*PPPPPPPP*/ARNBQKBNRA w KQkq - 0 1'

not the fen string you give above.

gaintpd:
Hello alwey, please test your current code with this fen string
`arnbqkbnra/*pppppppp*/a8*/*8*/*8*/*8a/*PPPPPPPP*/ARNBQKBNRA w KQkq - 0 1`
It will make you clear on this issue.

alwey:
OK, I see you are right. The FEN notation starts at the black side.

gaintpd:
Great! we eventually remove this bug that is hidden in the symmetric wall squares distribution.

alwey:
@gaintpd: Thank you for your help!
Now FEN` arnbqkbnra/*pppppppp*/*8*/*8*/*8*/*8A/*PPPPPPPP*/ARNBQKBNRA w KQkq - 0 1` is shown as

gustav3c

and flipped as

gustav3d

gaintpd:
@alwey Great! I think at this time the code can handle the flipped board and the asymmetric wall square distribution.

ilaripih:
Good stuff! And thanks @gaintpd for helping @alwey with testing and debugging :+1:

PR #589, opened 2020-07-10
Add support for r-mobility (https://wiki.chessdom.org/R-Mobility)

manyoso:
Introduce support for r-mobility (https://wiki.chessdom.org/R-Mobility).
    
    * Adds basic support for scoring via r-mobility
    * Outputs r-mobility score at end of game
    * Adds RMobilityResult tag to pgnout
    * Adds '-rmobility' flag to cutechess-cli which disables insufficient
      material adjudication
    * Add support for r-mobility scoring based on point system.

### DIFF
--- a/projects/lib/src/board/result.h
+++ b/projects/lib/src/board/result.h
@@ -22,6 +22,7 @@
 #include "side.h"
 #include <QMetaType>
 #include <QCoreApplication>
+#include <cmath>
 
 namespace Chess {
 
@@ -43,6 +44,14 @@ public:
 	void setResult(float r) { m_result = r; }
 	Side winning() const { return m_winning; }
 	void setWinning(Side side) { m_winning = side; }
+	float score(Side side) const
+	{
+		float winningScore = 0.5f + powf(0.5f, 1.0f + ceilf(m_result));
+		if (side == m_winning)
+			return winningScore;
+		else
+			return 1.0f - winningScore;
+	}
 	QString toString() const
 	{
 		if (!isValid())
--- a/projects/lib/src/board/result.h
+++ b/projects/lib/src/board/result.h
@@ -26,6 +26,47 @@
 namespace Chess {
 
 /*!
+ * \brief The r-mobility (https://wiki.chessdom.org/R-Mobility) result of a chess game
+ *
+ * The RMobilityResult class is used to store the rmobility result of a chess game,
+ * and to compare it to other results.
+ */
+class LIB_EXPORT RMobility
+{
+	Q_DECLARE_TR_FUNCTIONS(RMobility)
+public:
+	explicit RMobility() : m_isValid(false), m_result(0.0f), m_winning(Side::White) {}
+	explicit RMobility(float r, Side w) : m_isValid(true), m_result(r), m_winning(w) {}
+	bool isValid() const { return m_isValid; }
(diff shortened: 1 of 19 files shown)


alwey:
Interesting. Why are you tracking r-mobility for every game position under test?

manyoso:
> Interesting. Why are you tracking r-mobility for every game position under test?

This doesn't add r-mobility scores to test and check them, it rather just adds a default value so that tests of tournaments continues to work correctly.

alwey:
Hi. As far as I can see you only need one r-mobility outcome per game. This value is then transformed into a r-mobility floating point score. This score is added to the tournament statistics. I guess this is where your main interest is.  In addition the r-mobility game result is stored into a PGN-tag so that the r-mobility scores can also be calculated and processed externally.

Are these assumptions correct?

alwey:
> 
> 
>     Interesting. Why are you tracking r-mobility for every game position under test?
> 
> This doesn't add r-mobility scores to test and check them, it rather just adds a default value so that tests of tournaments continues to work correctly.

Ah - there has been a misunderstanding. I did not refer to the test suite but to all the positions that cutechess examines when a game is processed. These are many more positions than actually played within the game. Thus most of the r-mobility calculations are wasted. I think only one calculation per game is necessary.

manyoso:
> > ```
> > Interesting. Why are you tracking r-mobility for every game position under test?
> > ```
> > 
> > 
> > This doesn't add r-mobility scores to test and check them, it rather just adds a default value so that tests of tournaments continues to work correctly.
> 
> Ah - there has been a misunderstanding. I did not refer to the test suite but to all the positions that cutechess examines when a game is processed. These are many more positions than actually played within the game. Thus most of the r-mobility calculations are wasted. I think only one calculation per game is necessary.

This patch *only* calculates r-mobility for positions actually played during a game. It does not calculate it for positions not played or positions that are created only for purposes of checking legality for instance.

projects/lib/src/board/result.h
```diff
@@ -22,9 +22,59 @@
...
+ * \brief The r-mobility (https://wiki.chessdom.org/R-Mobility) result of a chess game
+ *
+ * The RMobilityResult class is used to store the rmobility result of a chess game,
+ * and to compare it to other results.
+ */
+class LIB_EXPORT RMobility
+{
+	Q_DECLARE_TR_FUNCTIONS(RMobility)
+public:
+	explicit RMobility() : m_isValid(false), m_result(0.0f), m_winning(Side::White) {}
```
ilaripih, line 39: Here also I'd like the implementations to be in the source file, and preferably separate header and source files for this class.

projects/lib/src/chessgame.cpp
```diff
@@ -267,7 +267,10 @@ void ChessGame::onMoveMade(const Chess::Move& move)
 
 	// Get the result before sending the move to the opponent
 	m_board->makeMove(move);
-	m_result = m_board->result();
+	const bool useRMobility = m_adjudicator.rMobilityEnabled();
```
ilaripih, line 270: Like @alwey noticed, do we have to calculate r-mobility after every move? Couldn't we do it only if the game has ended? And maybe that calculation could be done in `m_board->result()` function?

ilaripih (changes requested):
Thanks for the PR! This looks useful to some advanced users, especially if TCEC already uses it.

Issue #351, opened 2018-01-31
Few plausible handicap ideas 


Nordlandia:
Idea 1:  Option to allocate more time for engine 1 if engine 2 is considerable stronger.

Idea 2:  If engine 1 is weaker than engine 2, engine 1 can think during opponents turn while engine 2 can't ponder during his turn. 

Idea: 3: weaker engine play white only in maches. Already proposed in #321

The purpose of a handicap, is to equalize the strength between the engines and offer more fair conditions taken the strength difference into consideration.

alwey:
At least Option 1 should be feasible, actually I had thought about this. Time handicap can also be useful for characterisation of engines (e.g. in self play). Note: You can already use time handicaps with the CLI.

My opinion about idea 2 is not quite as favourable, just my EUR 0.02.

Idea 3 is a duplicate, as you already stated, of GUI improvement #321. Also, this option is available already in the CLI.

Nordlandia:
Few days ago one guy in _tcec chat_ claimed that 4-core with 2x time is ahead if facing 8-core with 50% less time and insisted that 4-core is still favorite if facing 8-core + ponder. 

In short: 

4-core with 2.0x time is ahead if facing 8-core with 0.50x base time
4-core is still favorite if 8-core is allowed to use ponder with half the time alloted.

I hoped idea number 2 was technically feasible.

alwey:
@Nordlandia :

> 4-core with 2.0x time is ahead if facing 8-core with [...] base time
> 4-core is still favorite if 8-core is allowed to use ponder with half the time alloted.


This is what I would expect and what I have seen when testing stockfish-8 with _cutechess-cli_ some months ago, in a little rr-tourney with one, two, and four cores having four, two and one amount of thinking time. The seach is basically a sequential algorithm. When executed in parallel, the efficiency drops.

Nordlandia:
Permanent brain handicap is interesting concept though.

Ipmanchess:
For engine testers is this also very interesting..would be a nice option!

Issue #161, opened 2016-10-01
Player-specific time controls and opening books


ilaripih:
Cutechess-cli already supports this. It should be possible to set time control and opening book (note: not opening suite) independently for each player in both NewGameDialog and NewTournamentDialog. By default those options should apply to all players like they do currently.

Note to self: this task is a lot bigger than it seems at first.

Nordlandia:
This related to my first handicap ideas #351

rwbc:
I would really like to see player-specific TCs in the very next release. It is very helpful currently in e.g. selftesting LCZero, if you just have a weak GPU. It helps simulating much better hardware and get more useful results.
Currently I am using cutechess-cli for this, but sometimes I wish to look at the board too.
Otherwise I would use WB of course, but there are some bugs in the timeodds implementation.

roy7:
@rwbc That's exactly what I'm doing. For instance, 1M nodes in LC0 and 200M nodes in stockfish.

ilaripih:
Changed the title because player-specific time controls are now possible in GUI, for both 1 vs 1 matches and tournaments.

PR #599, opened 2020-08-08
GUI: Option to use individual time controls in games and in tournaments

### LINKED ISSUE #413: Separate time controls/nodes per engine
Hi there. In the CLI version I can use something like nodes=X or tc=Y to have different settings for each engine. However in the GUI I can only set a single nodes or tc limit which is applied to both engines (like -both in CLI).

It'd be nice if there was a way to set these separately for each engine in the same game. (In my personal case I'm using lc0 with 1M nodes vs stockfish with 200M nodes.)

### LINKED ISSUE #619: Allow different time control (or node limits) for each player to enable time odds games
I'd like to run an engine tournament where all the players are the same engine, but with differing amounts of search time to see how the strength varies. As far as I can tell, this currently isn't possible because there is only a single time control option in the tournament settings window.

alwey:
This PR deals with two topics

1. Option to use split time controls for White and Black in individual games in the GUI

A new `TimeControlWidget` is introduced in order support the `TimeControlDialog`. It takes most of the dialog's components.
The dialog controls two `TimeControlWidgets`. The second `TimeControlWidget` (for Black) can be enabled on demand. By checking or unchecking a checkbox the dialog switches between common time controls for both sides or split time controls for White and for Black. 

`TimeControl timeControl2` is added to the `GameSettingsWidget` class, to save/read settings for the 2nd time control.
If used in the general settings (Tools->Settings->Games) the second time control is saved to the _cutechess.ini_ file.
It was necessary to add extra control logic for `hourglass` mode to prevent diverging settings for the sides.

2. Option to use individual time controls in tournaments in the GUI

A new context menu is added to the tournament players' list if some player(s) are selected. The menu item "Edit Time control" can be used to edit time controls individually. A simplified time control dialog is shown. It looks mostly like the old time control dialog, the 2nd time control is unused. Multiple-item selection in the player list is supported for time control updates, (Hint to users: also use the _SHIFT_ and the _CTRL_-modifiers when selecting for best results).

A `QVector` is added to `NewTournamentDialog` to store the individual time controls.

At tournament creation an individual time control is used when valid else tournament controls are used instead.
The hourglass mode is enabled or disabled uniformly according to the tournament's common settings.

Note:
In total, the `TimeControlDialog` is used in 4 ways:
1. General game settings ( via `GameSettingsWidget`, two time controls available, values are read from / saved to cutechess.ini).
2. Time controls for an individual game  (via `GameSettingsWidget`, two time controls available)
3. Common Time control of a new tournament (via `GameSettingsWidget`, but only one time control available)
4. Individual time control of a tournament player (via players' list context menu if player(s) selected: only one time control)

The PR relates to #161, the opening book part of #161 is still missing.
Resolves #413

I hope this is useful

**Update**: 
I attached some Screenshots below.
This PR also resolves #619, which came later. Also this addresses item 1 in request #351

### DIFF
--- a/projects/gui/src/gamesettingswidget.cpp
+++ b/projects/gui/src/gamesettingswidget.cpp
@@ -31,6 +31,7 @@
 GameSettingsWidget::GameSettingsWidget(QWidget *parent)
 	: QWidget(parent),
 	  ui(new Ui::GameSettingsWidget),
+	  m_splitTimeControls(false),
 	  m_board(nullptr),
 	  m_isValid(true)
 {
@@ -46,6 +47,9 @@ GameSettingsWidget::GameSettingsWidget(QWidget *parent)
 		this, SLOT(showTimeControlDialog()));
 	m_timeControl.setMovesPerTc(40);
 	m_timeControl.setTimePerTc(300000);
+	m_timeControl2.setMovesPerTc(40);
+	m_timeControl2.setTimePerTc(300000);
+
 	ui->m_timeControlBtn->setText(m_timeControl.toVerboseString());
 
 	connect(ui->m_browseOpeningSuiteBtn, &QPushButton::clicked, this, [=]()
@@ -127,6 +131,11 @@ QString GameSettingsWidget::chessVariant() const
 TimeControl GameSettingsWidget::timeControl() const
 {
 	return m_timeControl;
+}
+
+TimeControl GameSettingsWidget::timeControl2() const
+{
+	return m_timeControl2;
 }
 
 bool GameSettingsWidget::pondering() const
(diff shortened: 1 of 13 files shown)


gekkehenker:
Thanks for this commit.
Works great for me.

alwey:
Some screenshot to illustrate the features:

t4
_The time control for individual games (via Game->New) can now be set individually for White and Black._

t1
_The same goes for the general settings of time control (via Tools->Settings->Games)._

t3
_The checkbox in the top left corner switches between common time controls and split time controls [t2](https://user-images.githubusercontent.com/6425738/147849548-b0c603da-7f9d-4ed8-bd05-331e7f02fe9c.png)._


**Time controls can be set individually for tournament players.**
Also see the example with an engine playing multiple instances with different time controls given in #619.

t5
 _Right-clicking an entry and selecting "Edit Time Control" calls the time control dialog._

t2
 _In this example the configurations have been renamed by the user to reflect the individual time controls. The time control dialog for the last engine is shown._ 

t6
_The user should add the TC field to the resultformat, especially if the names of the tourney players are neutral._

ilaripih:
New PR where I continued this: https://github.com/cutechess/cutechess/pull/806
Closing this PR as duplicate

PR #605, opened 2020-08-13
add shako support

gaintpd:
Add shako support for cutechess. I have test the implementation using the fairy-stockfish engine several times. Every thing is OK as far as now.
The art design of the Cannon is from the default.svg file in cutechess.
The code for cannon move generation are copy directly from my xiangqi implementation.
The code for double move generation for  pawn in the initial position are copied from grandboard.

many thanks goes to ianfab for his kindly help on the explanation of the occupation of the King and Rook after castling.


alwey:
@gaintpd : I think is patch is of very good quality.

As this variant is from 1997, we have to be sure that we do not infringe patents or trademarks.

There are some formatting errors (whitespace, the project uses tabulators instead of space characters for indentations).

It is good to have only one commit for one topic. Here we have four commits, including three merge commits in between. But this is no problem because  `git cherry-pick`  works and picks one commit with the right stuff.

I try to avoid touching the base classes `Board` and `WesternBoard`. Not only to not clutter the classes, but also to avoid problems for standard chess or other variants. However, `WesternBoard::setCastlingTarget` is necessary for setting the target when the castling does not take place on the base rank. This method does not affect existing variants at all. This protected method should also have a short documentation in the header file.
An alternative would have been the introduction of a method `castlingRank()` in analogy to `castlingFile()`  but this is less direct.

You also included updates for the man page and the help text - thumbs up.
Test cases for this variant might be added to `projects/lib/tests/chessboard/tst_board.cpp` .

projects/lib/src/board/westernboard.h
```diff
@@ -201,7 +201,9 @@ class LIB_EXPORT WesternBoard : public Board
 		 * FEN string which succeed the en passant field.
 		 */
 		virtual QString vFenIncludeString(FenNotation notation) const;
-
```
alwey, line 204:
```
/*!
 * Set the castling target for \a side towards castling side \a cside to square index \a square.
 */
```

projects/lib/src/board/westernboard.cpp
```diff
@@ -851,7 +851,12 @@ void WesternBoard::setCastlingSquare(Side side,
 		xorKey(m_zobrist->castling(side, rs));
 	if (square != 0)
 		xorKey(m_zobrist->castling(side, square));
-	rs = square;
```
alwey, line 854: formatting, these 3 lines should not be patched

projects/lib/src/board/westernboard.cpp
```diff
@@ -851,7 +851,12 @@ void WesternBoard::setCastlingSquare(Side side,
...
+    rs = square;
+}
+
```
alwey, line 856: formatting

projects/cli/res/doc/help.txt
```diff
@@ -68,6 +68,7 @@ Options:
 			'racingkings': Racing Kings Chess
 			'rifle': Rifle Chess
 			'seirawan': S-Chess (Seirawan Chess)
+            'shako': Shako
```
alwey, line 71: use tabs instead

projects/lib/src/board/westernboard.h
alwey: Protected and public methods should have a comment above their declarations.

gbtami:
@alwey "As this variant is from 1997, we have to be sure that we do not infringe patents or trademarks."
Shako is available on pychess site and the author (Cazaux) never mentioned on our discord anything about patents or trademarks. (Contrary to Ed Trice :wink: )
If you want to ask him to be sure, you can find his e-mail at the bottom of http://jlg.cazaux.free.fr/

alwey:
@gbtami: Thank you very much for your help. I just sent an e-mail with our request to Mr. Cazaux.

gaintpd:
@alwey Thank you very much for your patient review. I will update the code based on your review.

> It is good to have only one commit for one topic. Here we have four commits, including three merge commits in between. But this is no problem because `git cherry-pick` works and picks one commit with the right stuff.

This is a problem that originated from the sync of my fork to the official repo. As far as now, I don't know how to avoid this problem, it is highly appreciated that you can offer me your suggestion on how to handle it.

> I try to avoid touching the base classes `Board` and `WesternBoard`. Not only to not clutter the classes, but also to avoid problems for standard chess or other variants. However, ` WesternBoard::setCastlingTarget` is necessary for setting the target when the castling does not take place on the base rank. This method does not affect existing variants at all. This protected method should also have a short documentation in the header file.

I cant agree more with your opinion.

> An alternative would have been the introduction of a method `castlingRank()` in analogy to `castlingFile()` but this is less direct.

This is a good advice, but I think the code can be refactored after it was merged into the master branch which make it possible to test it abundantly, and make this framework more flexible to handle more variants with the similar feature with **shako** in a graceful way.

> Test cases for this variant might be added to `projects/lib/tests/chessboard/tst_board.cpp `.

When I implement a new variant, I always play it as much as possible with the engines and myself, I did not take advantage of this facility. Your advice is very good, I will learn how to use it.

gaintpd:
@gbtami Thank you very much for the info you offered.

projects/lib/src/board/shakoboard.cpp
```diff
@@ -0,0 +1,230 @@
...
+            if(capture.isEmpty())
+            {
+                if(obstacle == 0)
+                    cannonRelOffsets.append(targetSquare - sourceSquare);
+            }
+            else
+            {
+                obstacle++;
+                if(obstacle == 2 && capture.side() != side)
+                {
```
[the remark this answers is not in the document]
gaintpd: I have modify the code as your request.

alwey:
Great - we got permission from Jean-Louis Cazaux.
He is positive, even suggesting to consider his 12x12 variant _Metamachy_.
I will forward the correspondence for the record of this project.

@gaintpd : Thank you, I will give a more detailed answer below.

projects/lib/src/board/westernboard.cpp
gaintpd: This is strange, I have never change the code before.

projects/lib/src/board/westernboard.cpp
gaintpd: Thank you

projects/lib/src/board/westernboard.h
gaintpd: Thank you, I have add the comment above the declaration.

gaintpd:
> Great - we got permission from Jean-Louis Cazaux.
He is positive, even suggesting to consider his 12x12 variant Metamachy.
I will forward the correspondence for the record of this project.

I am willing to impement **Metamachy** variant if I can get access to the rule specification if everything is possible.

projects/cli/res/doc/help.txt
gaintpd: Thank you

gaintpd:
@alwey I have change the code as your request, but it is strange that the 3 line of code was changed above the `setCastlingTarget` function in the source file `westernboard.cpp`. Thank you very much for your help to review my code.

gbtami:
@gaintpd see http://history.chess.free.fr/metamachy.htm :)

alwey:
@gaintpd ; I think there are still other format errors, but let's work on the single commit topic first.
Now we have an additional commit, `a1d0b0e remove the format error`.
The commit `5ba21f3 add shako support` is the main commit.

In your local repository check out the shako branch and make a copy by reference:
```
git checkout shako
git checkout -b shako.backup
```
Fetch the newest master from github
`git fetch upstream master`  (step A)
 (you might have another name for the cutechess main repository)

Have a look at master to check all is well: 
`git log master`

Now rebase shako onto the new master
`git rebase master shako`
`git log shako`
There should be only two commits above master now.

To get them into a single commit use
`git rebase --interactive master shako` (step B)
A list of commits is shown, the main commit "add shako support" should be the first.
Change the word "pick" before "remove the format error" into "f" or "fixup".
Press CTRL-K X or whatever your active editor needs to save the file.

`git log `
should show only one commit above master.

If you are done with your work and satisfied you should force-push your branch to github to override your previous commits.
`git push -f mygithub shako.` (step C)

The results will apperar in this PR.
Strictly speaking, only steps A,  B and C are necessary.

gaintpd:
@gbtami Thank you very much for your info. I have read the rule of **Metamachy**, I think the difficult part is the mirror move at the start phase just like the piece selecting phase in **Musketeer** variant.

gaintpd:
@alwey Thank you very much for your detailed instruction. Now I have reconstructed the shako branch in my local repo and have done a forced push to my fork.
As I use QtCreator to edit the code. It seems whatever I tried to remove the format error, some format error still exist like a ghost, which is annoying. What's the reccommand configuration of the code editor to edit the cutechess source code. I think the format error is originated from the configuration of the code editor.

alwey:
@gaintpd : Although QtCreator is installed on my computer, I have not really used it. I am using KDevelop for cutechess, because I used it for a Qt project years ago. The indents are by tabulators, 8 characters each.

gaintpd:
@alwey Thank you for your reply. Maybe I also need to switch to KDevelop to editor the code to keep insistent with the code style, as far as now, I cannot find a solution in the QtCreator IDE to keep consistent with the requirede code style.

I have replace the `setCastlingTarget` with `castlingRank` method as the result of your review, and I have changed the code in a new commit **ede4253** to record your creative method purposely.

Thank you very very much for the help in this patch.

alwey:
@gaintpd : Hello, I think there has been a misunderstanding /wrt castling. Actually, I found your old solution with `setCastlingTarget` better than using `castlingRank`, because it is more versatile and direct. I would suggest dropping commit `ede4253`.

> An alternative would have been the introduction of a method castlingRank() in analogy to castlingFile() but this is less direct.

My bad, I should have left the comment out to avoid confusion.

For castlingRank I would suggest `int WesternBoard::castlingRank() {return 0; //first rank } `WesternBoard::castlingRank() {return 1; //second rank } and only calculate the target squares in  `WesternBoard::vInitialize()`. It would be easier to use. 

At least I would suggest subtracting rank offset 2 from the current expressions in the `castlingRank` implementations in `WesternBoard` and `ShakoBoard` and add this value 2 to the calculations in WesternBoard::vInitialize(), because this reflects the shift between rank numbers on the usable squares vs all array squares.

Sorry for making things complicated.

gaintpd:
@alwey  It is not your bad, Your `castlingRank` approach is better than mine. And above all, your approach make the code quality better compared to my ininital commit. I will  suggest ilaripih to merge the commit ede4253 to the master(main) branch. And you do make the code more graceful in my opinion.

gekkehenker:
I wonder if my love for Terachess is known to the author? 

Great work on this commit btw.

PR #622, opened 2020-09-13
Scale chess engine timeouts

### LINKED ISSUE #459: winboard "feature done" bug
It looks like cutechess-cli hangs with winboard engine's that take to too long to initialize -- which in my case is due to loading neural networks. The winboard protocol says this:

> done (integer, no default)
> If you set done=1 during the initial two-second timeout after xboard sends you the "xboard" command, the timeout will end and xboard will not look for any more feature commands before starting normal operation. If you set done=0, the initial timeout is increased to one hour; in this case, you must set done=1 before xboard will enter normal operation.

So I can do a "feature done 0" ... then time-taking operation ... "feature done 1".
This would allow for initialization times of upto 1 hour but cutechess-cli times out despite what its code suggests

`
	else if (name == "done")
	{
		write("accepted done", Unbuffered);
		m_initTimer->stop();
		
		if (val == "1")
			initialize();
		return;
	}
`

This has caused severe problems at TCEC.

More at http://talkchess.com/forum3/viewtopic.php?f=2&t=68956

Daniel

### LINKED ISSUE #618: Unable to run Stockfish 12 with a large amount of cache
## Output of `cutechess-cli --version` / `cutechess --version`
cutechess-cli 1.2.0
Using Qt version 5.11.1
## How did you run Cute Chess exactly?
GUI
## Expected behavior
SF runs
##Actual behavior
SF never starts.
What you observed.
When setting SF's cache to more than around 20 GB, it takes too long to allocate RAM on startup and CuteChess stops it because it doesn't get the ping response. It should recognize this or allow disabling ping.
## Steps to reproduce the behavior
Set a lot of cache in SF settings

alwey:
Some issues have been raised regarding cutechess's strict timing especially at game start. Problems have been reported both for NN engines and for configurations allocating vast amounts of Hash memory.

To get around this a new CLI engine option `tscale` is introduced to scale the protocol timeouts. It can be used on a per-player (per-configuration) basis, e.g.
`cutechess-cli -engine cmd=engine1 proto=uci tscale=2.0 -engine cmd= ...` .

To achieve this a scale factor for timeouts is added to the `ChessEngine` and `EngineConfiguration` classes. The relevant engine protocol timeouts will all be multiplied with this factor. This apllies to ping, quit, idle, and  protocol start timeouts. Also for CECP engines the init timeout is scaled.

In the GUI a spin box for the scaling factor has been added to the engine configuration dialog.
The value of 1.00 is the default and displayed in gray. The increase/decrease arrows of the spin box have intentionally been omitted to discourage tinkering with the timeouts.

The scale factor can have values between 0.2 and 120.0.
I hope this is useful.

Resolves #459
Resolves #618

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -481,6 +481,10 @@ Enable pondering if the engine supports it.
 Set the search depth limit.
 .It Ic nodes Ns = Ns Ar count
 Set the node count limit.
+.It Ic tscale Ns = Ns Ar factor
+Scale engine timeouts by
+.Ar factor .
+Only use this option if necessary.
 .El
 .Sh EXAMPLES
 Play ten games between two Sloppy engines with a time control of 40
--- a/projects/cli/res/doc/help.txt
+++ b/projects/cli/res/doc/help.txt
@@ -215,5 +215,7 @@ Engine options:
   nodes=N		Set the node count limit to N nodes
   ponder		Enable pondering if the engine supports it. By default
 			pondering is disabled.
+  tscale=FACTOR		Scale engine timeouts by FACTOR. Only use this option
+			if necessary.
   option.OPTION=VALUE	Set custom option OPTION to value VALUE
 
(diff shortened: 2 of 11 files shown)


ilaripih:
I changed "timeout_scale_factor" to "timeoutScaleFactor" for consistency; otherwise looks good. Merged!

PR #625, opened 2020-09-14
Add 1/2 pseudocounts to WDL in SPRT.

### LINKED ISSUE #360: SPRT with big ELO differences
I was trying to run an experiment using a random play engine (that avoids randomly winning) as an Elo reference. Of course that engine never wins. That prevents the SPRT tests to stop. I understand that you cannot compute the Elo of an engine that always lose, but that shouldn't be a problem for finding the probability of its Elo being outside a threshold. An engine with 100% wins that doesn't pass an SPRT test doesn't sound like it should be the expected behaviour.

kiudee:
As mentioned in #360, the SPRT class is currently not able to handle the case where one or more counts (WDL) is 0.
This simple fix implements a Dirichlet([0.5, 0.5, 0.5]) prior on the parameters of the trinomial distribution, which corresponds to [Jeffreys prior](https://en.wikipedia.org/wiki/Jeffreys_prior) (also see [[1]](https://en.wikipedia.org/wiki/Beta_distribution#Jeffreys'_prior_probability_(Beta(1/2,1/2)_for_a_Bernoulli_or_for_a_binomial_distribution))). This specific prior has many interesting properties the most important of which for our application is that it maximizes the mutual information between the sample and the parameters.

Fixes #360 

# References
[1] https://en.wikipedia.org/wiki/Beta_distribution#Jeffreys'_prior_probability_(Beta(1/2,1/2)_for_a_Bernoulli_or_for_a_binomial_distribution)


alwey:
I guess the signature of 
`SprtProbability::SprtProbability(int wins, int losses, int draws)`
should be changed to
 `SprtProbability::SprtProbability(double wins, double losses, double draws)`.

kiudee:
Indeed, I pushed the change.

alwey:
Don't forget the header-file ;-)

kiudee:
At some point I should properly set up my development environment here on the windows workstation. ;-)
I pushed a fix to the SPRT unit test which now works for the new regularized values.

alwey:
I think the value in question should be  -4.69.

kiudee:
> 
> 
> I think the value in question should be -4.69.
> This value should be calculated independently, e.g. by using a calculator {program}.

Yes, I am going through my manual calculations to make sure the results are correct.

kiudee:
I re-did the calculation and the correct result is:
```
-4.690026035584188023926644139716793351289185573436562465480648057...
```

kiudee:
Thanks for the active code review - looks like the pull request is ready for review now.

alwey:
You are welcome.

I would like to suggest using `const double wins` etc. in sprt.cpp line 170ff. instead of `const float`. These variables will be converted to double anyway.
From my own experience I assume the maintainers prefer a single commit for this PR.

kiudee:
Thanks for the hint. Will squash it down into one commit.

kiudee:
I added a comment to explain why 0.5 is added.

respel:
Hi, I was testing this PR and it seems like the original issue of the engine not passing the sprt test still exists.

kiudee:
@respel Isn’t your issue resulting because you did not provide the number of rounds?

This pull request fixes the issue that you won’t be able to terminate until you have seen at least one of each possible game outcome (WDL). I don’t see how it is related to #636. Maybe you can elaborate?

respel:
@kiudee I see. That was the issue. Thanks for helping out!

PR #626, opened 2020-09-19
Support Makpong  (Makruk variant)

alwey:
Makpong is a variant of Makruk where the King may not move out of check.  The King may capture an attacking piece if in range, but cannot capture out of double check.

 This variant is designed to reduce draws. Over the board it is used as tie breaker in Makruk single elimination tournaments.

`MakpongBoard` modifies the move generation of `MakrukBoard` and the adjudication /wrt to material. It has been tested with Fairy-Stockfish.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -133,6 +133,8 @@ Loop Chess (Drop Chess Variant)
 Los Alamos Chess
 .It losers
 Loser's Chess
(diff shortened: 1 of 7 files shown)


gekkehenker:
Thank you for adding this variant.

While I don't like Makruk, this one actually seems like a great variant. Simple rules, low drawrate, low opening bias.

PR #629, opened 2020-10-18
Support Lines of Action (LOA)

alwey:
[_Lines of Action_](https://en.wikipedia.org/wiki/Lines_of_Action) is an abstract strategy game on a 8 x 8 board. It has been introduced by Claude Soucie (–1960).

At start both sides have 12 pieces lined up at the boundary squares of the board. The corner squares are free. Black has their pieces on the 1st and 8th rank of the board and starts the game. White occupies the
a- and the h-file and moves second. Both sides take turns moving.

A piece can move horizontally, vertically, or diagonally. The step size  of a move equals the total number of pieces on the line of movement. Opponent pieces can be captured but not be jumped over. Own pieces can be jumped over but not be captured.

The goal of the game is to connect all of a side's on-board pieces so that they are touching horizontally, vertically, or diagonally.

When all pieces of both sides get connected by the same move the game is either drawn (Sackson's draw rule) or the side making the last move wins (Soucie's rule - which is used here). From Wikipedia: _"Despite the intention of the inventor of LOA, most present day tournaments including the World Championships at the Mind Sports Olympiad score simultaneous connection as a draw. "_ Therefore the implementation also supports Sackson's draw rule, but currently the controlling flag is set to false.

If a side has no legal moves left this implementation adjudicates a win for the opponent. When a position occurs for the third time the game is drawn.

This implementation is based on `WesternBoard`, it. uses the standard chess implementation and chess notation.

loa0 
_LOA Starting Position_
loa1
_All black pieces are connected, Black wins._ 

There is a variant "Scrambled Eggs" where white and black pieces are placed alternately on the outer squares.
 FEN: 1PpPpPp1/p6P/P6p/p6P/P6p/p6P/P6p/1pPpPpP1 b - - 0 1
This variant can be set-up via FEN (directly or per openings file). 

loa3
_"Scambled Eggs" starting position_

I hope this is OK to be included.

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -127,6 +127,8 @@ Kinglet Chess
 King of the Hill Chess
 .It knightmate
 Knightmate
+.It linesofaction
+Lines of Action
 .It loop
 Loop Chess (Drop Chess Variant)
 .It losalamos
--- a/projects/cli/res/doc/help.txt
+++ b/projects/cli/res/doc/help.txt
@@ -56,6 +56,7 @@ Options:
 			'kinglet': Kinglet Chess
 			'kingofthehill': King of the Hill Chess
 			'knightmate': Knightmate
+			'linesofaction': Lines of Action
 			'loop': Loop Chess (Drop Chess)
 			'losalamos': Los Alamos Chess
 			'losers': Loser's Chess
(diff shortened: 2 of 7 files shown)


alwey:
rebased and replaced `QRegExp` by `QRegularExpression`

PR #647, opened 2021-01-03
Maintainance

alwey:
This PR contains small fixes.

- replace the old option `-both` in the manual page by `-each`.
- if a game is drawn after a user defined maximal move number has been reached then set a corresponding result text to be more specific.


artoj:
Merged, thanks!

PR #653, opened 2021-02-21
Fix ai-wok variant name in docs

ianfab:
The cutechess CLI expects to use the variant name `ai-wok`, while the docs say `aiwok`. This PR should make the naming consistent.


artoj:
Merged, thanks!

PR #686, opened 2021-11-12
Rounding node numbers.

zz4032:
Rounding "NPS" and "Nodes" to numbers with 3 significant digits:

screenshot


alwey:
"Nodes" is exact information. I think it should be displayed as is. Node speed is a different matter. Its accuracy depends on the elapsed move time.

zz4032:
In my opinion there is no benefit for the user to see the exact number of calculated nodes of a running engine in a GUI.
For exact speed measurements benchmarks are more appropriate.

PR #700, opened 2022-01-01
GUI: Add tooltip text to resultformat edit

alwey:
In the CLI help to control the format of the result table can be listed by `-resultformat help`.
This patch adds a tooltip with available format fields and shortcuts to the GUI's format line edit.


ilaripih (changes requested):
Any text displayed to the user should support localization, i.e. should be wrapped in `tr(english_string_here)`. Then it's possible to use Qt Linguist to provide full translations for other languages.

Also, the amount of code is high enough that I think we should have a separate function to construct the tooltip text.

PR #707, opened 2022-01-22
Re-use previous tournament's settings

### LINKED ISSUE #209: Tournament settings not remembered
It would be great if when creating another tournament that we wouldn't have to re-enter all the same settings again like PGN output file, Opening Suite, Depth of book, re-adding the engines again and changing its parameters if you don't want to use the global settings for that particular run.  It'd be nice if it would use the settings and engine parameters from last time.

Also, it'd save us a lot of working in modifying the saved pgn file if when stopping an unfinished tournament the games with no result wouldn't be saved into the pgn. The way it is now if I start a 200 game tournament but I stop it after a 100 (or whatever)  then the pgn will still contain the unfinished game with a "no result" after each game. Makes it hard to post clean pgn's online.

Lastly, the results. I find it horrible that it doesn't list a nice cross-table that also lists the number of wins, losses, and draws for each engine instead of just showing us the elo, games, score%, and draws%

alwey:
This patch changes the usage of the `NewTournamentDialog` in the GUI. Currently this widget is deleted after use (OK or Cancel) and a new instance with the default settings is used for the next tournament. 

This code re-uses the existing `NewTournamentDialog` with old values to speed up the set-up of the next - probably similar - tournament. A new _Reset_-Button triggers the destruction of the dialog. After a short while a new dialog appears with user's default values.

The method `NewTournamentDialog::readSettings` has been simplified and corrected.

This PR is an **alternative** to the patch which uses the old dialog with values from _cutechess.ini_ after reset, PR #706.

Resolves #209

nt1
_A "Reset"-button has been added to NewTournamentDialog._

_Example: First dialog shown before ..._
nt4

_... and a new dialog appears after pressing the "Reset"-button._
nt5

HTH.


EveannChessEngine:
How can I get and executable with this feature ?

eduherminio:
In case it helps someone else, binaries are available in [v1.4.1-eduherminio-707](https://github.com/eduherminio/cutechess/releases/tag/v1.4.1-eduherminio-707) and [v1.3.2-eduherminio-707](https://github.com/eduherminio/cutechess/releases/tag/v1.3.2-eduherminio-707).

PR #716, opened 2022-02-12
Option to control debugging on UCI engines

### LINKED ISSUE #505: cutechess does not send "debug on"
when running cutechess with -debug , at no point does it send "debug on" to the engine. Is this intentional? 

Thanks

alwey:
This pull request adds support for protocol based debugging for UCI engines. In the CLI debugging can be switched on individually per engine by adding `debug` to the `-engine` section, and generally (only active for eligible engines) by adding `debug` to the `-each` section. This can also be accomplished by adding the parameter `all` to the `-debug` option:
`-debug all` is equivalent to `-debug -each debug`. As before the `-debug` option alone only switches on cutechess's own debugging: It shows the communication with the engines and displays all engine input and output.

In the GUI a new checkbox in `EngineConfigurationDialog`  is used to individually enable UCI engine debugging.

d01

This PR adds the methods `debugEnabled` and `setDebugEnabled` to `EngineConfiguration` and introduces the formal engine option "debug" e.g. 
`cutechess-cli -engine debug option.op1 ...`  or `cutechess-cli -each ponder debug ...` .

The new method `ChessEngine::debugEnabled()` is added.  `UciEngine::startGame` sends "debug on" if` debugEnabled()`.
The Xboard protocol does not support debugging in the same way.

Add support for GUI:
Add `m_debugCheck` to `EngineConfigDlg` and supporting logic to `EngineConfigurationDlg`.

Add support for CLI:
Add option argument "all", so `-debug all` is equivalent to `-debug -each debug`.

Resolves #505

### DIFF
--- a/docs/cutechess-cli.6
+++ b/docs/cutechess-cli.6
@@ -295,8 +295,11 @@ games.
 Set the interval for printing outcomes to
 .Ar n
 games.
-.It Fl debug
+.It Fl debug Bo Cm all Bc
 Display all engine input and output.
+In addition, all eligible engines will be switched to debug mode if argument
+.Cm all
+is given.
 .It Fl openings Cm file Ns = Ns Ar file Cm format Ns = Ns Bo Cm epd | Cm pgn Ns Bc Cm order Ns = Ns Bo Cm random | Cm sequential Bc Cm plies Ns = Ns Ar plies Cm start Ns = Ns Ar start Cm policy Ns = Ns Bo Cm default | Cm encounter | Cm round Bc
 Pick game openings from
 .Ar file .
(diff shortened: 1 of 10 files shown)


alwey:
Sorry for the late reply. I think _Komodo Chess_ is an exemplary implementation of the standards, e.g. _Komodo 14_.

ilaripih:
I think this is otherwise fine, but the `-debug all` CLI option is not really needed as `-each debug` works well and then there's less documentation and parser changes needed. So I'm going to push a slightly modified version of this change.

ilaripih:
Alright, the changes (slightly modified) are in. I tested the debug engine config with Komodo 1.

PR #717, opened 2022-02-13
Adjudicate draw in case of a timeout if the opponent has no means to win

### LINKED ISSUE #710: Wrong lose on time if insufficient mating material
In this position:
`8/8/8/4k3/4p3/4N3/4K3/8 w - - 0 1`
if black run out of time it's a draw because white has insufficient mating material, while cutechess give white a win.

alwey:
A game is drawn in case of a timeout if the opponent has no means to give checkmate by any regular sequence of moves (Ref.: FIDE rule 6.9). Cutechess currently adjudicates a loss on timeout anyway.

This patch resolves simple cases of lone kings. It does not detect dead positions like impasse by pawn blockade.

`ChessPlayer::forfeit` calls a new virtual method `Board::winPossible(Side)` to check whether the opponent has any pieces beside their King. If not, then the game is drawn.

Resolves #710

### DIFF
--- a/projects/lib/src/board/board.cpp
+++ b/projects/lib/src/board/board.cpp
@@ -920,4 +920,9 @@ Result Board::tablebaseResult(unsigned int* dtm) const
 	return Result();
 }
 
+bool Board::winPossible(Chess::Side) const
+{
+	return true;
+}
+
 } // namespace Chess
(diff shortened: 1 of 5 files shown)


ilaripih:
Looking good

PR #723, opened 2022-05-19
add initial cmake support

artoj:
This commit includes initial support for CMake. Most things do not work,
such as tests, but it's possible to build the gui, lib and cli projects
and generate working executables.

QMake will be deprecated in the future so at some point we need to make
the transition to CMake, which is practically the only choice forward.
CMake is used to build Qt 6.

In its current form QMake and CMake exist in paraller: the project can
be build in either two systems but the plan is to remove QMake support
completely.

Changes to the source tree were intended to be kept as minimal as
possible to accommodate CMake.


artoj:
To build Cute Chess with cmake with this PR:

```
cd <to the cutechess root directory>
mkdir build && cd build
cmake ..
cmake --build .
```

ilaripih:
Looks good, and great effort! Is the Qt::SkipEmptyParts error here relevant to this PR: https://lgtm.com/projects/g/cutechess/cutechess/logs/rev/pr-fcaf2e84c57447d73f8efda96bb9d1449155d2c2/lang:cpp/stage:Build%20merge_06c7735cab90831a33fb9017db8086e1f57e1615

artoj:
The CMake build now contains Windows, Linux and MacOS support, and unit tests. Default CI has been switched to CMake. If no comments arise, I'll merge this in a few days.

artoj:
CMake support has been merged and default CI action is to build with CMake. The possibility to build Cute Chess with QMake still exists but will be removed in the near future.

PR #727, opened 2022-06-11
cmake: provide `install` target

artoj:
Provide an `install` target with CMake.

I'm unsure whether the `engines.json` manual should be renamed to `cutechess-engines.json`.


artoj:
Example usage:
```
mkdir build && cd build
cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local
make
sudo make install
```
After the install step these should work:
```
cutechess
cutechess-cli
man cutechess-cli
man engines.json
```
It should be possible to launch Cute Chess using an application launcher.

artoj:
No negative comments, so let's get this in.

Issue #722, opened 2022-05-18
Improve the release process


artoj:
Currently the releasing Cute Chess takes too many manual steps, is error prone and slow.

In order to make releases more often, we need to greatly improve the release process.

- [x] Releases are done by GitHub Actions --- not by manual process
- [x] Releases are triggered by a Git push. This can be for example creation of new tag, such as "v1.3"
- [x] Release artifacts are uploaded with the GitHub release
- [x] Linux releases are done with AppImage --- not by packaging individual distro formats such as DEB or RPM. This should streamline creation of Linux binaries. Native packages should be done by the distribution packaging teams.
- [x] Improve support for native distribution packing by providing proper `make install` target.
- [x] Windows releases are initially done by creating a Zip archive with the binaries
- [x] Once we have Linux AppImage and Windows Zip, introduce Windows installer with Inno Setup
- [ ] macOS binary .app creation

artoj:
CMake support has been merged which was prerequisite for the improvements listed above.

Next I'm going to improve the situation for native Linux distribution packaging by providing a proper `make install` target.

artoj:
CMake now has a basic `install` target. Next up: generation of AppImages. Once we have that in, I think we can start providing binary snapshots of the current `master` branch.

artoj:
There's now AppImage creation in the `release.yml` workflow which creates a release when a tag is pushed.

artoj:
We have now Linux AppImage and WIndows zip release generation working. Next up: Windows installer

artoj:
We have now Windows installer generation.

artoj:
I consider this one. MacOS image generation will be after 1.3.0

PR #734, opened 2022-08-18
Nice arrow

kdave:
Draw nicer move arrow, similar to lichess.org arrows. Second patch has images for preview, not intended for merge.

### DIFF
--- a/projects/gui/src/boardview/boardscene.cpp
+++ b/projects/gui/src/boardview/boardscene.cpp
@@ -529,20 +529,32 @@ void BoardScene::addMoveArrow(const QPointF& sourcePos,
 	if (!m_settings.value("ui/show_move_arrows", true).toBool())
 		return;
 
-	QLineF l1(sourcePos, targetPos);
-	QLineF l2(l1.normalVector());
+	QPolygonF arrow;
+	QLineF origline(sourcePos, targetPos);
 
-	l1.setLength(l1.length() - s_squareSize / 2.5);
-	l2.setLength(s_squareSize / 3.0);
-	l2.translate(l2.dx() / -2.0, l2.dy() / -2.0);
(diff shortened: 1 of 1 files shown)


artoj:
Looks good to me.

Ping @ilaripih

ilaripih:
Agreed, looks good. Can you remove the "Add arrow previews" commit and I'll merge this?

kdave:
Sure, commit removed and pushed.

tissatussa:
So you are going to change the look of the yellow move arrow .. can there also be an option (in Settings / General) to en/disable the arrow ? That would be nice. I think other people had this request also ..

kdave:
There is a setting: Tools -> Settings -> General -> Show move arrows [x]

kdave:
Although, what's common on web chess servers is that the last move is only denoted by hilighted squares, the arrows are user-drawn. I think people are used to this and I'd like to bring it to cutechess too but I realize this may be intrusive and breaking the current visual style.

tissatussa:
@kdave !? I'm using Cute Chess v1.2.0 on Xubuntu 22.04 and i see no such option ..

CuteChess_settings

kdave:
I see, it's a new option added after 1.2.0, I was running it from git.

tissatussa:
> I see, it's a new option added after 1.2.0, I was running it from git.

@kdave !? how can that be? v1.2.0 is the latest version ..

artoj:
There has been a lot work in Git since v1.2.0. We haven't released 1.3.0 because we have problems in our release process (outlined in https://github.com/cutechess/cutechess/issues/722) which I've been fixing for the past months. We're slowly getting there but not yet.

tissatussa:
@artoj respect .. i have no problems with 1.2.0

artoj:
Merged, thanks!

PR #768, opened 2023-07-29
QStringView based Xboard protocol parsing

artoj:


artoj:
Needs testing that these changes do not break anything.

ilaripih:
The idea here is to remove dependency on `QStringRef` which is no longer including in the core Qt 6 module and has been moved to Qt 5 Core Compatibility APIs: https://doc.qt.io/qt-6/qstringref.html

This PR does a lot of that work but does not completely remove that dependency.

PR #802, opened 2024-07-19
build Windows releases with Qt 6

artoj:


EveannChessEngine:
Still not save/load tournaments. This is a MUST for intensive testers..

PR #808, opened 2024-07-22
fix cases of negative time when timemargin is > 0

gahtan-syarif:
currently when timemargin/expirymargin is set, there can be cases where the updated time left can be negative. this is of course undefined behavior for engines. as such, with this pr, it resets the time to 0 before increment is applied when the time goes negative but hasn't exceeded the expiry margin.

fix https://github.com/cutechess/cutechess/issues/809

### DIFF
--- a/projects/lib/src/timecontrol.cpp
+++ b/projects/lib/src/timecontrol.cpp
@@ -367,6 +367,10 @@ void TimeControl::update(bool applyIncrement)
 	else
 	{
 	        int newTimeLeft = m_timeLeft - m_lastMoveTime;
+
+		if (newTimeLeft < 0)
+			newTimeLeft = 0;
+		
 		if (applyIncrement)
 			newTimeLeft += m_increment;
 		setTimeLeft(newTimeLeft);


ilaripih:
Hi, and thanks for the PR! The code looks good, but I'll still do some testing tomorrow.

alwey:
I think, resetting negative times to zero would be mathematically incorrect and help cheating. The engines have to deal with negative rest time plus increment.

gahtan-syarif:
> I think, resetting negative times to zero would be mathematically incorrect and help cheating. The engines have to deal with negative rest time plus increment.

This could lead to the program sending the engine negative time in the next move if the increment is lower than the time exceeded which should never happen under any circumstance. This behavior caused problems in chess.com's ccc tournament not long ago. I think it makes more sense to treat timemargin as a buffer in which if an engine exceeds the specified time within the timemargin then it is treated as if the engine never exceeded the time in the first place, so in this case setting it to 0 makes sense

ilaripih:
> I think, resetting negative times to zero would be mathematically incorrect and help cheating. The engines have to deal with negative rest time plus increment.

I was worried about the same thing. Even with this change, the engine going below 0 into the time margin would still have the "stolen" time reduced from their following moves (e.g. 40 moves per 5 minutes, and we're at the end of the first batch of 40 moves). And even without this change, in "X time per move" time controls, engines could steal time margin amount of time for each move.

I think the only way this proposed change could be exploited is in "X time per N moves" time control, with or without increment, if the engine goes below zero before the last move of a batch of N moves:

1. Let's say the time control is 40 moves per 5 minutes and time margin 200ms.
2. The engine uses all their time in the first 10 moves.
3. For the next 30 moves, the engine makes their moves within the time margin and stays in the game. Basically the engine now has an increment of 200ms per move, except that any leftover time in the margin is not added to their time.

This would be impossible in the current implementation, as it should - the engine would not be able to make 30 moves with 0 time, unless the time margin was absolutely massive.

So how do you (@gahtan-syarif and @alwey) feel about this proposal:

- We change nothing in `lib/TimeControl`, i.e. the above exploit would not be possible
- When sending the remaining time left to the engine, it's floored to 0 so even if the engine has negative time left in our internal time keeping, we tell them they have zero time. Or if the engines cannot handle receiving 0 as time left, maybe we could use 1 centisecond, i.e. the shortest amount of time in the XBoard protocol

If the only issue with the current time margin implementation is that engines don't know what to do when they receive negative time left, this would solve it without changing our internal time keeping.

gahtan-syarif:
I see, if keeping seperate time between the internal bookkeeping and cutechess time is an option, then i suggest this: for every move, the reported time to the engine is `max((timeLeft - timemargin), 1)`. This way there wont be negative values in both cutechess's timeleft and the engine reported time @ilaripih

alwey:
> *  We change nothing in lib/TimeControl, i.e. the above exploit would not be possible
> * When sending the remaining time left to the engine, it's floored to 0 so even if the engine has negative time left in our internal time keeping, we tell them they have zero time. Or if the engines cannot handle receiving 0 as time left, maybe we could use 1 centisecond, i.e. the shortest amount of time in the XBoard protocol

Yes, so the internal calculation does not change, only the communication to the engine. I think it is fine.

@gahtan-syarif : I feel the reported time left should be as close as possible to the actual time left, so I would prefer to modify the transmitted time only in the case of negative times.

gahtan-syarif:
> Yes, so the internal calculation does not change, only the communication to the engine. I think it is fine.
> 
> @gahtan-syarif : I feel the reported time left should be as close as possible to the actual time left, so I would prefer to modify the transmitted time only in the case of negative times.

i feel if this were to only be applied during negative time then you get a lot of cases where the engine receives a lot of `go wtime 1`'s nea the end and the time allocation per move would be awry (not ideal for engine testing). its best to inform the engine that it has a reduced time for search due to overhead right from the getgo so that the engine can plan its timecontrol ideally

gahtan-syarif:
since resetting the time to 0 in lib/timecontrol seems to not be wanted, ill go ahead and close this pull request

PR #823, opened 2024-09-15
Fix error in standard algebraic notation disambiguity logic

spinojara:
In the position "8/1k5p/4P3/1q3b2/6PP/7K/5QR1/1q1q4 b - - 0 53" the move b5d3 is Q5d3+ not Qb5d3+.


ilaripih:
Thanks, this looks correct. I'll add some test automation for the SAN formatting and will merge this when I'm confident that nothing breaks.

Issue #817, opened 2024-08-23
cutechess did not end game with 3-fold repetition due to pinned en passant position


gahtan-syarif:
```
[Variant "From Position"]
[FEN "6k1/1p2p1rp/rP1pR3/2pP1pP1/p1P2P1P/R5K1/8/8 b - - 0 1"]

1... h5 2. Rh6 Rh7 3. Re6 Rg7 4. Rh6 Rh7 5. Re6 Rg7
```
5.Rg7 should have been ended as a 3-fold draw which cutechess did not seem to notice. this is because it incorrectly thinks 1.h5 is an en passant position (which it isnt because the ep capture is illegal). these types of cases happen very rarely but i think is worth fixing nonetheless as the above position did happen during a chess engine sprt.

alwey:
I think you found a problem.

Indeed cutechess gives the FEN after h5 as `6k1/1p2p1r1/rP1pR3/2pP1pPp/p1P2P1P/R5K1/8/8 w - h6 0 2`
This is a peculiarity of the notation. It represents a double step made by the correspending pawn but does not reflect the (il)legality of such an e.p. capture.

Adding a legality check the modified FEN would be  `6k1/1p2p1r1/rP1pR3/2pP1pPp/p1P2P1P/R5K1/8/8 - - 0 2`, because the pawn on g5 cannot leave its file.

Cutechess checks its list of known position keys (zobrist hash keys) where also the e.p. status is embedded to detect repetitions. The legality of the e.p. capture is not taken into account when modifying the zobrist hash.

```
void WesternBoard::setEnpassantSquare(int square, int target)
{

	m_enpassantTarget = square ? target : 0;
	if (square == m_enpassantSquare)
		return;

	if (m_enpassantSquare != 0)
		xorKey(m_zobrist->enpassant(m_enpassantSquare));
	if (square != 0)
		xorKey(m_zobrist->enpassant(square));

	m_enpassantSquare = square;
}

```

PR #828, opened 2024-12-23
Fix illegal en passant

JeremieHuteau:
This addresses https://github.com/cutechess/cutechess/issues/817

Adding some kind of `isEnPassantLegalAfterDoublePawnPush` function that would be called in `vMakeMove` is not reliable due to variants (I've run into issues with Horde and Grid so far, but I expect there would be others).

The general solution (which would work for any property dependent on a future move's legality) is to make the double pawn push entirely, then run a check for all pseudo-legal EP moves available and reset the the en passant square if none is legal.

This PR introduces a new flag to `Board` (`needsFutureDependentUpdate`) that is set inside `WesternBoard::vMakeMove` when a double pawn push with consequent pseudo-legal en passant is made. `Board::makeMove` then calls `WesternBoard::vApplyFutureDependentUpdate` which does the en passant legality check and resets the en passant square if no en passant move is legal.

### DIFF
--- a/projects/lib/src/board/board.cpp
+++ b/projects/lib/src/board/board.cpp
@@ -710,6 +710,22 @@ void Board::makeMove(const Move& move, BoardTransition* transition)
 	xorKey(m_zobrist->side());
 	m_side = m_side.opposite();
 	m_moveHistory << md;
+
+	if (m_needsFutureDependentUpdate)
+	{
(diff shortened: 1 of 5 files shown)


ilaripih:
Thank you @ArcheCadia! I can confirm that your fix works. I ended up implementing this a little bit differently (see referenced commit), but your implementation and the test cases were a great base.

PR #845, opened 2025-09-25
Migrated from QRegExp to QRegularExpression on ./porjects/lib/src/engineprocess_win.cpp. Resolves #396.

### LINKED ISSUE #396: Migrate from QRegExp to QRegularExpression
Qt 5.0 introduced `QRegularExpression` class that has a better API, is faster and more compliant with other regex implementations. We should migrate from the old `QRegExp` class to the newer one.

iskandarem:
This PR replaces the last remaining usage of `QRegExp` in `projects/lib/src/engineprocess_win.cpp` with `QRegularExpression`.

- `#include<QRegExp>` changed to `#include<QRegularExpression`.
- Object of  `QRegularExpression` created instead of `QRegExp`.

Please let me know if I missed anything or if improvements are needed.


artoj:
I'll close and re-open the PR to allow the workflows to run.

artoj (changes requested):
The `QRegExp` and `QRegularExpression` classes have different APIs thus the build is failing on Windows. It's not enough to do string replacement, the API calls need to be updated as well.

PR #859, opened 2026-06-03
docs: document optional value for -debug option

### LINKED ISSUE #842: cutechess-cli -debug requires a value which the docs don't mention
## Steps to reproduce
```
cutechess-cli -debug
```
I'm on commit 4b331698

## Expected behavior
Either cutechess-cli should support -debug with no value, or the docs should say that a value is required. currently all the info I can find is "-debug: Display all engine input and output", implying that debug _shouldn't_ have a value.

I'd like to fix this myself as its far from a big issue but I'm not sure what the process is for editing the docs (i.e. what I actually need to edit to change all the docs and help text).

## Actual behavior
```
Warning: Empty value for option "-debug"
```

The option actually needs to be set as "-debug all"

re2zero:
## Summary
- Fix the help text and man page to document that `-debug` accepts an optional value `[all]`

## Changes
| File | Change |
|------|--------|
| `projects/cli/res/doc/help.txt` | `-debug` → `-debug [all]` |
| `docs/cutechess-cli.6.txt` | `-debug` → `-debug [all]` |

## Context
The `-debug` option activates debug mode to display engine I/O. When passed `all` as a value, it also enables the per-protocol debug option for each engine. The documentation only said "Display all engine input and output" without mentioning the optional `[all]` argument, causing user confusion.

## Test Plan
- [x] Verified that code (`main.cpp:278`) declares `-debug` as `QVariant::String` with maxOccurrences=1
- [x] Verified that code (`main.cpp:486`) checks for `value == "all"` to enable per-engine debug
- [x] Documentation now matches the actual code behavior

Fixes #842

### DIFF
--- a/docs/cutechess-cli.6.txt
+++ b/docs/cutechess-cli.6.txt
@@ -181,7 +181,7 @@ DESCRIPTION
      -outcomeinterval n
 	     Set the interval for printing outcomes to n games.
 
-     -debug  Display all engine input and output.
+     -debug [all]  Display all engine input and output.
 
      -openings file=file format=[epd | pgn] order=[random | sequential]
 	     plies=plies start=start policy=[default | encounter | round]
--- a/docs/cutechess-cli.6.txt
+++ b/docs/cutechess-cli.6.txt
@@ -181,7 +181,8 @@ DESCRIPTION
      -outcomeinterval n
 	     Set the interval for printing outcomes to n games.
 
-     -debug [all]  Display all engine input and output.
+     -debug [all]  Display all engine input and output.  Passing the
+                  value all also enables per-engine protocol debugging.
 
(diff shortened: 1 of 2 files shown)


ilaripih (changes requested):
Thanks for this PR! I think the docs should also explain what the "all" value does, as you've already explained in the PR description.

re2zero:
Good point, thanks for the review! I've updated the docs to explain that passing `all` also enables per-engine protocol debugging.

matt-dot-net:
I agree with updating the docs, but I think the implementation should also be adjusted so the flag works without a parameter when it’s intended to be optional. The current behavior effectively requires an argument even though the option is described as optional.

I’ve also opened a related PR that makes the flag work as a bare switch, so the implementation matches the optional-argument behavior discussed here.

PR #869, opened 2026-06-13
fix pricipal variation parsing in xboardengine.cpp

ilaripih:
This also fixes the comparison operators in MoveEvaluation, and the xboardengine tests.


ilaripih:
The `line.trimmed()` change fixes the parsing of thinking output for engines like Crafty that start their output with a whitespace character.

PR #871, opened 2026-06-14
use ChessEngine::tokenize in UciEngine

ilaripih:
This commit replaces the QStringRef based tokenizer functions in UciEngine with the QStringView based ChessEngine::tokenize function.


ilaripih:
Can't get rid of Qt5 Compatibility yet because we still have a dependency on QRegExp: https://doc.qt.io/qt-6/qtcore5compat-module.html
But this removed QStringRef completely from the project.

PR #876, opened 2026-07-02
fix #866 - use QProcess for engine communication also on Windows

### LINKED ISSUE #866: Maybe replace custom engineprocess_win with QProcess
The origin story for creating our QProcess for Windows was high latency when reading output from the engine. This made engines time out when using any reasonably fast time controls. Since then there's been a lot of development for `qprocess_win.cpp` in Qt. It's possible that the latency issue was fixed already in https://github.com/qt/qtbase/commit/dcdab683d6bfb6b0ad81b5903236ae70a71246e4

**Acceptance criteria**

- Test fast time control games (e.g. 1-3 seconds per game) on Windows with latest Qt, using Qt's QProcess instead of CuteChess' custom `engineprocess_win`. Test both XBoard and UCI engines.
- If everything goes well in the tests, and we don't have latency-induced timeouts, get rid of `qprocess_win` and `pipereader_win`.

ilaripih:
Upstream Qt has finally fixed the latency issues with the Windows implementation of QProcess.


.github/workflows/release.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           contents: write
         strategy:
           matrix:
-            qt_version: [6.8.3]
+            qt_version: [6.9.3]
```
artoj: I'm wondering should be use version 6.9 only on MacOS?

.github/workflows/release.yml
```diff
@@ -100,7 +99,7 @@ jobs:
           ISDIR: 'C:\Program Files (x86)\Inno Setup 6'
         strategy:
           matrix:
-            qt_version: [6.8.3]
+            qt_version: [6.9.3]
```
artoj: Ditto

artoj:
* `README.md` needs to be updated: still mentions qt5compat

artoj:
* `CHANGELOG.md` should probably mention that QProcess is now in use instead of our own implementation

ilaripih:
Code review fixes applied.

PR #878, opened 2026-07-09
gui: work around macOS 27 beta QMacStyle not drawing check option checkboxes

drewster99:
### Summary
On **macOS 27.0 beta 3 (26A5378j)**, boolean (`check`-type) engine options render as **blank, un-clickable cells** in the engine options table — the Value column shows nothing and the option can't be toggled.

### Cause
`QMacStyle` no longer paints the item-view check indicator for a `QTreeView`/`QTableView` cell on this beta (Apple reworked AppKit controls on top of SwiftUI; the same path also *crashes* QMacStyle when drawing an `NSSlider`). It renders correctly on **macOS 26.6**, under **Fusion**, and in a `QListView` — so it's a QMacStyle regression against the beta, not a cutechess bug.

Minimal pure-Qt reproduction (no cutechess):
```cpp
#include <QApplication>
#include <QTreeView>
#include <QStandardItemModel>
int main(int argc, char **argv) {
    QApplication app(argc, argv);
    QStandardItemModel model(1, 1);
    model.setData(model.index(0, 0), Qt::Checked, Qt::CheckStateRole);
    QTreeView view; view.setModel(&model); view.show();
    return app.exec();
}
```
| Environment | Result |
|---|---|
| macOS 27.0 beta 3, QMacStyle | ❌ blank |
| macOS 26.6, QMacStyle | ✅ checkbox |
| any macOS, `-style fusion` | ✅ checkbox |

### Change
Draw the checkbox and handle its toggling ourselves in `EngineOptionDelegate`, gated behind `#ifdef Q_OS_MACOS`. `paint()` sizes/positions the indicator from `SE_ItemViewItemCheckIndicator` and blits it from an origin-anchored pixmap (QMacStyle ignores the option rect for `PE_IndicatorCheckBox`); `editorEvent()` toggles `CheckStateRole` on a left-click of the indicator or Space/Select. The model is unchanged, so non-macOS builds are unaffected.

### Notes
- This is a **workaround for a beta-only regression** — Apple or Qt may fix it before macOS 27 ships, making it redundant (but harmless: it draws the same checkbox via the style).
- The guard is compile-time `Q_OS_MACOS`, so the custom drawing also runs on macOS versions where QMacStyle works (26.x, and presumably the 27 release). If preferred, I can scope it more tightly (runtime check / affected builds only), or hold until 27 ships — happy to adjust.

### DIFF
--- a/projects/gui/src/engineoptiondelegate.cpp
+++ b/projects/gui/src/engineoptiondelegate.cpp
@@ -25,6 +25,31 @@
 #include <QtDebug>
 #include "pathlineedit.h"
 
+#ifdef Q_OS_MACOS
+#include <QApplication>
+#include <QStyle>
+#include <QStyleOptionButton>
+#include <QPainter>
+#include <QPixmap>
+#include <QMouseEvent>
+#include <QKeyEvent>
+#include <QAbstractItemModel>
+
+namespace
+{
+// Rect of the check indicator within a value cell. Ask the style for the exact
+// rect it uses for item-view check indicators so the size and vertical
+// centering match everything else. Expects an option already passed through
+// initStyleOption() (so HasCheckIndicator and widget are set).
+QRect macCheckIndicatorRect(const QStyleOptionViewItem& option)
+{
+	QStyle* style = option.widget ? option.widget->style() : QApplication::style();
+	return style->subElementRect(QStyle::SE_ItemViewItemCheckIndicator,
+				     &option, option.widget);
+}
+} // anonymous namespace
+#endif
+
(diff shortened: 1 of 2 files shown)


ilaripih:
Thanks for the PR! Can you rebase this to latest master, where the Qt version on mac was bumped up? And can you try what happens with the latest Qt 6.11 version?

I'm a bit hesitant to merge something that's likely going to be fixed soon in a new version of Qt - but then again this change is well contained and would be easy to revert at any time.

drewster99:
agreed - it might become irrelevant, possibly quickly.

I'll take care of rebasing and rechecking when I get back tomorrow.

PR #879, opened 2026-07-12
Replace remaining QRegExp usage

### LINKED ISSUE #396: Migrate from QRegExp to QRegularExpression

aouxwoux:
Replaces the final `QRegExp` use in the Windows engine-process command-line parser with `QRegularExpression::globalMatch()`. Each iterator result contributes its full captured match, preserving the existing tokenization behavior while removing the obsolete API dependency.

Validation:
- `git diff --check`
- Confirmed no `QRegExp` references remain under `projects/`
- Native build/tests not run locally because CMake and Qt are unavailable; the repository CI builds Windows, Linux, and macOS with Qt 6.8.

Fixes #396.

### DIFF
--- a/projects/lib/src/engineprocess_win.cpp
+++ b/projects/lib/src/engineprocess_win.cpp
@@ -19,7 +19,7 @@
 #include "engineprocess_win.h"
 #include <QDir>
 #include <QFile>
-#include <QRegExp>
+#include <QRegularExpression>
 #include <QMutexLocker>
 #include "pipereader_win.h"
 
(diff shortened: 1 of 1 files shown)


ilaripih:
Thank you for the PR! I'm happy to say that changes to EngineProcess are no longer needed though, because we got rid of EngineProcess completely: https://github.com/cutechess/cutechess/pull/876

PR #880, opened 2026-07-14
fix: use full moves instead of plies in parsing and exporting of PGNs

### LINKED ISSUE #873: PGN annotations show plies not full moves
## Output of `cutechess-cli --version`

```
cutechess-cli 1.5.1
Using Qt version 6.11.1
Running on macOS Sequoia (15.6)/arm64
```

## How did you run Cute Chess exactly?

```bash
cutechess-cli \
  -engine cmd=./engine name=Engine1 \
  -engine cmd=./engine name=Engine2 \
  -each proto=uci tc=2 \
  -openings file=mate.epd format=epd \
  -games 1 \
  -pgnout match.pgn
```

Where `mate.epd` contains: `5k2/8/8/8/8/8/6Q1/4R2K w - - 0 1` (it has mate in two).

## Expected behaviour

I think the move annotations should show the mate distance in full moves. At least, it would make more sense for it to be consistent with UCI, as my engine that I was testing with Cute Chess was correctly outputting `info score mate 2` and thus the PGN should have read:

```
1. Qb7 {+M2/5 0.050s} Kg8 {-M1/5 0.050s} 2. Re8# {+M1/5 0.050s}
```

## Actual behaviour

The mate distance is shown in plies.

```
1. Qb7 {+M3/5 0.050s} Kg8 {-M2/5 0.050s} 2. Re8# {+M1/5 0.050s}
```

## Root cause

Not sure if this is the only instance of this issue, but in `moveevaluation.cpp`, `MoveEvaluation::scoreText()` recovers the internal ply count and appends it directly to `"M"` without dividing back to full moves.

mkutay:
This is especially important for "mate in x". Previously, the exported PGN would output it through plies, instead of full moves as in the specification.

In addition, when we are parsing a PGN, we also convert the full moves to the ply count.

Fixes #873

### DIFF
--- a/projects/lib/src/moveevaluation.cpp
+++ b/projects/lib/src/moveevaluation.cpp
@@ -137,7 +137,7 @@ QString MoveEvaluation::scoreText() const
 		{
 			if (m_score < 0)
 				str += "-";
-			str += "M" + QString::number(absScore);
+			str += "M" + QString::number((absScore + 1) / 2);
 		}
 		else
 			str += QString::number(double(m_score) / 100.0, 'f', 2);
--- a/projects/lib/src/pgngame.cpp
+++ b/projects/lib/src/pgngame.cpp
@@ -574,10 +574,19 @@ QMap< int, int > PgnGame::extractScores() const
 		bool isMateScore = s.contains('M');
 		if (isMateScore)
 			s.remove('M');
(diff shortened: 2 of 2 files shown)


mkutay:
Note that I haven't added any tests, as I'm not so sure how I would approach it! But I would be more than happy to learn.

